PR #1029, opened 2012-07-03
Diff for merge requests is broken with submodules

funglaub:
This is a quick fix to enable diffs in merge requests when your repository contains submodules.


PascalBrouwers:
This feature is awesome! Can it also be applied to composer modules?

PR #1248, opened 2012-08-17
Omniauth Support 

funglaub:
Hey guys. 

First of all, thanks for the great app!

I added Omniauth support for Google-OAuth2, Twitter and GIthub to gitlab.

It works like that:
A user can choose between pre-configured omniauth providers to login along with his username/password. This can be found in the 'Authentications' (the former 'Password') tab. 

Configuration is done in config/gitlab.yml, including app_id and app_secret for each provider. The list of providers can easily be expanded by adding the needed gems to the Gemfile and providing a corresponding logo. The current logos were taken from http://github.com/intridea/authbuttons.

Omniauth can also be disabled in config/gitlab.yml, which is the current default setting.

I tried to integrate all the new functionality as good as possible into existing views.

Please, tell me what you think!


strk:
+1 if this will let users use own OpenID

PR #1426, opened 2012-09-10
Cucumber => Spinach

NARKOZ:
:sparkles:


alwinmark:
Why did you move away from cucumber? For me spinach seems to have no advantages and it lacks of features to really reuse steps.

So at the and those steps are clustered in features files and steps (like in cucumber) without the ability to reuse them as they are to dedicated to one use case.

So either you are completely out of mind our I'm missing your point.

alwinmark:
Also its very strange how you are using Gherkin:

You are never using `When` Statements... just `Given` and `Then`. You should only use `Given` if you are preparing the test and `Then` if you're checking the expecting behaviour. If a user is doing the actual tasks which are interesting for the test they should be wrapped with `When` Statements.

It's much clearer what a test should do, when it is failing if these conditions are clear.

vsizov:
Actually we are moving from spinach to rspec :) This is issue is outdated.

PR #3592, opened 2013-04-14
Implemented code search feature

jacargentina:
#### The feature in action

TypeSearchTerm
Type your search term on any project place...

CodeResultsReady
Then get your results with links to them!


dellasavia:
It would be interesting if can be possible search text into files...

PR #4684, opened 2013-07-31
Add support for REMOTE_USER omniauth strategy

annagrram:
Requested in issue: https://github.com/gitlabhq/gitlabhq/issues/481


Schniz:
This simple patch worked for us when we upgraded to 7.2 (from 6.0.1) with a simple diff.

sous-marin:
If I've well understood, with your patch, autologin can be done natively in GitLab. Do you have any instruction about the configuration in GitLab side ? @annagrram   Thanks (I'm in Omnibus 10)

PR #4706, opened 2013-08-05
Added the necessary items to make PAM authentication work.

dwa012:
I have added what is needed to for PAM authentication to work. I am including the wiki instructions as will. There are a few steps to do outside of this package to configure the system to make this type of authentication to work.
### Install Process
- Have the libpam0g-dev package installed
- Install the rpam gem
  
  `gem install rpam`
- If on Ubuntu, you will need to replace the unix_chkpwd binary.[https://github.com/dwa012/pam-1.1.3-ubuntu-git](https://github.com/dwa012/pam-1.1.3-ubuntu-git). Follow the instructions in the readme. This will authorize the git user along with the root user to check passwords against the shadow file. You will have to weigh your own security concerns to determine if you want to enable PAM. You can also use the generated deb package to install the binary. If using the binary you will want to freeze that package, so that is will not be replaced in a future upgrade.
- Uncomment the PAM omniauth option in the gitlab.yml. You will want to modify the email argument to fit your needs. This will allow users whom are created at first login to have a valid email.


orodbhen:
It seems that the correct way to handle the permissions problem would be to use polkit, not by altering the PAM module.

PR #4810, opened 2013-08-19
CSV previewing

Floppy:
This is WIP code that will allow useful viewing and diffing of CSV files within Gitlab. It uses [coopyhx](https://github.com/paulfitz/coopyhx) to show the diffs.

Parts that still need work:
- [x] use same viewer for both file and diff view, for consistency
- [ ] make Blob#to_csv more robust to different CSV variants
- [ ] move Blob extensions into gitlab_git and github/linguist gems
- [x] improve naming of things in Javascript code - currently stolen straight from demo page


kolokd:
gitlab still doesn't have this feature. what a world we live in!
this is one  of the reason where gitlab sucks. another one is poor search result.

acxz:
Related work: https://gitlab.com/gitlab-org/gitlab/-/issues/14174

PR #5280, opened 2013-10-06
Add MathJax support in Wiki and Markdown files.

harbulot:
This patch enables [MathJax](http://www.mathjax.org/) rendering in wiki pages and in `.md` files.

MathJax needs to be installed separately. The [MathJax CDN](http://docs.mathjax.org/en/latest/start.html#mathjax-cdn) can be used, for example.

To enable this feature, simply set the MathJax URL (e.g. `http://.../MathJax.js?config=TeX-AMS-MML_HTMLorMML`) as the `mathjax_url` option in the `extra_config` section of `config/gitlab.yml`.

Note that, unlike other Markdown processors, the GitLab Markdown processor already processes `^`, so this symbol has to be escaped if used in a formula: `\^`.

Here are a couple of examples that can be used in wiki or markdown content:

```
Here is a mathematical formula:

$$f(x) = \sum\_{i=1}\^{N}{x\^2}$$

Here is an inline formula: \\( f(x) = \sum\_{i=1}\^{N}{x\^2} \\)
```

The formulae should be turned into mathematical notation after the page is loaded. The first time MathJax is used, this transformation might seem a bit slow, mainly because it needs to download a number of fonts. This improves once MathJax is cached.

### DIFF
--- a/app/views/layouts/_mathjax.html.haml
+++ b/app/views/layouts/_mathjax.html.haml
@@ -4,6 +4,7 @@
       url: '#{extra_config.mathjax_url}',
       dataType: 'script',
       crossDomain: true,
+      cache: true,
       success: function() {
         $('.wiki').each(function(k, item) {
           MathJax.Hub.Queue(["Typeset",MathJax.Hub,item]);
--- a/app/views/layouts/_mathjax.html.haml
+++ b/app/views/layouts/_mathjax.html.haml
@@ -1,0 +1,8 @@
+:javascript
+  $(document).ready(function() {
+    $.getScript('#{extra_config.mathjax_url}', function() {
+      $('.wiki').each(function(k, item) {
+        MathJax.Hub.Queue(["Typeset",MathJax.Hub,item]);
+      });
+    });
+  });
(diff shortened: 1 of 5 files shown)


DaAwesomeP:
:+1:

JoKeyser:
:+1: Really would be nice to have this feature. And I don't see the problem with this pull request (other than it's outdated). Are you really saying that 5 lines of code are "too much code duplication"?

DouweM:
I agree that there is not too much code duplication, but to be able to accept this merge request we would need:

- Code that works with the latest GitLab, everywhere where Markdown is rendered

- The feature enabled by default, loading MathJax from the MathJax CDN (https://docs.mathjax.org/en/v2.6-latest/start.html), with an appropriate configuration (TeX-MML-AM_CHTML is probably too large)

- Ideally some way to _not_ load MathJax if the Markdown doesn’t contain $$ delimiters. We could a Markdown processing filter that adds a class to the markdown body if it contains MathJax, and only load the script then.

- Reasonable equation delimiters, there seem to be some problems with the discussion in https://github.com/gitlabhq/gitlabhq/pull/5280. No idea if all of that’s still relevant.
- A config option in the admin area to disable MathJax (like “Gravatar enabled")

- Documentation on how to use and disable the feature

dosire:
Thanks for the comment @DouweM

PR #5378, opened 2013-10-20
Viewing IPython notebooks in GitLab

martijnvermaat:
Dear GitLab maintainers,

With this pull request I intend to get some feedback on extending GitLab with live inline rendering of IPython Notebooks.
## Introduction and rationale

The [IPython Notebook](http://ipython.org/notebook.html) is a web-based interactive environment for editing documents containing both executable code and rich text. The platform is becoming increasingly popular in the scientific community amongst others.

These documents are serialized as JSON and can be converted to HTML, PDF, and other formats using the IPython nbconvert tool. Static HTML renderings can also be shared over the web using the [IPython Notebook Viewer](http://nbviewer.ipython.org/) and indeed this is quite popular for viewing notebooks stored in GitHub.

However, due to GitLab's focus being more on repositories with controlled access, it is not possible to use the IPython Notebook Viewer service with notebooks stored in GitLab. Even if it were possible, this would bring unwanted security issues since everything on the Notebook Viewer is public.
## Implementation

This pull request is an ad-hoc implementation of live inline rendering of notebooks in GitLab. Clicking a notebook file in the GitLab web interface will convert the notebook to HTML using the IPython nbconvert tool and include it in the GitLab blob view as a base64 encoded data URI for an IFrame. The obvious benefit here is that everything happens inside the GitLab access controlled view of the blob.

[You can find some more documentation on the patch here](https://gist.github.com/martijnvermaat/6926070/).

Since this is only a proof-of-concept, I can point out some obvious areas for improvement:
- Somehow cache the nbconvert results.
- Test (and presumably fix) files (in the same repository) linked from the notebook.
- Add unit tests.
## Request for feedback

I think this would be a great feature for IPython Notebook and GitLab users and it would give GitLab a clear advantage over GitHub for hosting IPython Notebook, especially in private repositories.

Are you interested in a feature like this? If so, let's discuss necessary improvements. If not, I'll try to keep my own branch up-to-date (and still welcome any feedback!).

### DIFF
--- a/app/assets/stylesheets/generic/files.scss
+++ b/app/assets/stylesheets/generic/files.scss
@@ -64,6 +64,18 @@
       }
     }
 
+    &.ipython-notebook {
+      iframe {
+        width: 100%;
+        /* It is quite hard to choose a good height for the IFrame since we
+           cannot relate it to its contents (it's cross-domain/protocol). We
+           choose to have it very high, so we usually don't get nesting of
+           scrollbars (GitLab page and the IFrame). */
+        height: 16000px;
+        border: none;
+      }
+    }
+
     &.blob_file {
 
     }
--- a/app/helpers/tree_helper.rb
+++ b/app/helpers/tree_helper.rb
@@ -51,6 +51,52 @@ module TreeHelper
     filename =~ /^README(.txt)?$/i
   end
 
+  def ipython_notebook?(filename)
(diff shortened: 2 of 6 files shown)


davidovitch:
For the record, I created a corresponding feature request on Gitlab.com: [#1594](https://gitlab.com/gitlab-org/gitlab-ce/issues/1594)

PR #5760, opened 2013-12-04
Fix HTTP push to public repos

dermoth:
When doing an HTTP push, git (as of v1.7.9) first do an info/refs
request, and only if this request requires authentication it asks the
user for its password and authenticates further requests.

The initial request normally clears without auth on public repos as it
doesn't update any ref. This patch forces every git-receive-pack to
provide authentication.

### DIFF
--- a/lib/gitlab/backend/grack_auth.rb
+++ b/lib/gitlab/backend/grack_auth.rb
@@ -92,6 +92,9 @@ module Grack
           return false unless can?(user, action, project)
         end
 
+        # Never let git-receive-pack trough unauthenticated; it's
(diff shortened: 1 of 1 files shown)


detly:
I've been trying to write a script using [Dulwich](https://github.com/jelmer/dulwich) to manipulate my private Gitlab-hosted repo. As far as I could tell, using a private access token should work by having a header of the form `Private-token: whatever` in the request, and not needing a username/password. However, when accessing eg. `myproject.git/info/refs?service=git-receive-pack` I get a 401 back using this method.

Am I doing something wrong, or is returning `false` if the user is `nil` incorrect because it prevents access via private token?

PR #5958, opened 2013-12-30
 Added support for Go's repository retrieval.

brunoga:
- Simply adds the relevant meta link in the project's show action.
- Information about this meta tag can be found here:
  http://golang.org/cmd/go/#hdr-Remote_import_paths

Note I am not sure how to do unit tests for this, but I am willing to do those if someone can point me to how to do it. Also I did not update the CHANGELOG as I am not sure where to put the change. Is it under v6.5.0?

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -201,7 +201,7 @@ class Project < ActiveRecord::Base
     [Gitlab.config.gitlab.url, path_with_namespace].join("/")
   end
 
-  def web_url_without_scheme
+  def web_url_without_protocol
     web_url.split("://")[1]
   end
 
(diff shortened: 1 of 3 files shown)


marciocarmona:
I'm facing the same problem @lfyzjck had. But in my case I can't change my repo visibility to public.
I guess maybe GitLab should return the HTML with the meta tag when it receives the go-get=1 parameter instead of returning a 302 to the login page, or just starting accepting and redirecting the URLs adding the ".git" extension automatically

marciocarmona:
Apparently, #7693 will fix that! :smile:

PR #692, opened 2012-04-18
GitLab Localization (en, de)

alexdo:
Hey there!

Over the last days, I localized GitLab into English and German. To accomplish this, I made use of the Rails L18n API.
Commit ff3166be is a squashed commit of 7b0d606b .. bd514eda in my master branch. It includes all changes in the view as well as YAML files for **en + de**.

I know, the YAMLs are quite dirty. But in fact, GitLab is a rapidly evolving application. Over the last 10 days, I were really busy merging new commits from gitlabhq:master and adjusting to changes, thus refactoring of YAML structure should be another piece of work or localization would take ages to complete.

After inspecting all the views, I also noticed a inconsistency of labels(either "Save" or "Add a new abc..") and alert dialogs (DELETE). Sometimes only "Are you sure?", sometimes a more concise "Are you sure you want to ....?".

GitLab is a great piece of software and I'd really like to see it being localized into serveral languages. Keep up the good work, you're awesome!

PR #6455, opened 2014-03-02
Internationalization support

estrai:
Hi,
as per #692, this is an approach to add internationalization support to GitLab.
Not all pages are yet translated, most notably admin and help pages, but this can be done later.
I'm willing to maintain translations is a separate repository and, as it was suggested, use Transifex to collect translations.
As to concerns that this could slow down development I think this won't be the case if only English version will be supported, additionally it will simplify scenarios where a helper method was used to repeat the same message.


socketpair:
Please explain (in short) current status of supporting lcalisation in GitLab.

app/controllers/groups_controller.rb
```diff
@@ -131,7 +131,7 @@ def authorize_admin_group!
   end
 
   def set_title
-    @title = 'New Group'
+    @title = I18n.t("gitlab.groups.new_group")
```
socketpair: Why not to use https://github.com/ruby-gettext/gettext ? it have "contexts" support built in.

connorshea:
@socketpair https://gitlab.com/gitlab-org/gitlab-ce/issues/4012 

See that thread for the current status of internationalization of GitLab. We're probably going to go with gettext, but the process hasn't started yet and isn't currently a major priority for a number of reasons.

Please comment there if you'd like to discuss it further.

PR #6573, opened 2014-03-20
Added custom service url for Slack integration.

razielgn:
This addresses what's been discussed in #6550.
I'm still waiting to hear from Slack about the custom endpoint, so we should wait until they say it's good to go. :)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -133,7 +133,7 @@ gem "gitlab-flowdock-git-hook", "~> 0.4.2"
 gem "gemnasium-gitlab-service", "~> 0.2"
 
 # Slack integration
-gem "slack-notifier", "~> 0.2.0"
+gem "slack-notifier", "~> 0.3.0"
 
 # d3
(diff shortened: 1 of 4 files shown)


nazrhyn:
Back when this integration wasn't working, I wrote a node service to handle this for our team. It takes a slightly different approach to the notifications than the current integration does (e.g. the way the notifications are formatted and the fact that we just don't display issue modification notifications at all [which could address @Turbo87's concern]). It's been evolving since then and I recently cleaned it up for posting to GitHub.

The channels are handled with configuration settings that map the project ID to a Slack channel. Any project ID without a mapping goes to whatever the default is for the Slack webhook. Since this mapping exists, we have a single Slack webhook and use the same handler (in our case, `http://127.0.0.1:port`) for every project's webhook registration.

Here's the link, in case anyone's interested in some options: https://github.com/nazrhyn/gitlab-slack

PR #6912, opened 2014-05-02
Added support for Markdown formatting

ameily:
- Switched from using gfm() method to markdown(), which eventually calls gfm()
  - Previous fix (pull request 6777) caused a call stack overflow on master


cdbennett:
If you don't want to change commit message rendering unconditionally, at least make it a user-configurable option on a project-by-project or systemwide basis.  Note, the markdown.md file in GitLab CE still says that markdown is supported in commit messages. https://gitlab.com/gitlab-org/gitlab-ce/issues/1582

PR #7038, opened 2014-05-29
Adds commits calendar with ajax request to user profiles

drleviste:
Calendar of commits in user profiles

When user has commits, it will show a calendar covering a timeframe from the year before up to his latest commit
fdf6190a-e577-11e3-8af6-afc3f15771d8

When user has no commits, it will show a timeframe up to the current month
fe7f355a-e577-11e3-9ded-ff902e828c24

@randx this is from #6958 we, @kbleabres and I, added ajax request and edited some lines

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -144,6 +144,9 @@ gem "slack-notifier", "~> 0.3.2"
 # d3
 gem "d3_rails", "~> 3.1.4"
 
+#cal-heatmap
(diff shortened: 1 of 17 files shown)


123Haynes:
I've resolved the merge conflicts for this. But the gitlab functions have changed quite a bit so it doesn't work at the moment.
If someone wants to help, please take a look at this:
https://gitlab.com/haynes/gitlab-ce/commits/commit_calendar

PR #7518, opened 2014-08-16
Set charset encoding to UTF-8 for snippets

jvanbaarsen:
**What does this MR do?**
It makes sure that a raw snippet gets the HTTP headers charset=utf-8. That way snippets with weird characters gets rendered properly

**What issues does this fix?**
Fixes #2678


TorstenC:
Example: https://gist.github.com/TorstenC/362a97d2c74a6b522fd6
This Gist (snippet) is: Content-Type: text/plain; charset=windows-1252

@Dmitry: Why is this closed?

PR #7611, opened 2014-08-26
Diff headers made sticky.

yglukhov:
This MR allows diff file headers to become sticky (persistent). The reason for this MR is to simplify reviewing of a batch of files, when reviewer can forget the file, he's currently on.


yglukhov:
The problems are solved [here](#8760)

PR #7615, opened 2014-08-27
Validation of 'Create branch' and 'Create tag' buttons

yglukhov:
Fixed validation of "Create branch" button. The issue reproduced on Safari 7.0.6. The button was enabled on the newly loaded page with empty text-fields, and after clicking it it became permanently disabled, so that text-field modification didn't help.

Now it is reflecting current status of the text-fields.


yglukhov:
Thank you!

PR #7681, opened 2014-09-03
Extend user add api to allow adding users like from admin panel

duk3luk3:
- Add force_random_password parameter to have password of new user set
  randomly and expired
- Update docs
- Update tests

Rationale:
- When adding a user from the admin panel (Admin Area -> New user), the user is created with a random, expired password and in the welcome email, is directed to (re)set their password
- The user creation API call right now only allows for a password to be set that is supplied by the API caller
- Creating a user with a random expired password from the API is a desireable feature for e.g. batch-adding users without handling password distribution
- This PR adds that feature

This is a remake of my [old PR](https://github.com/gitlabhq/gitlabhq/pull/4805) that was abandoned since things about how user registration/signup works were expected to change soon. Since it's been a year and the PR is still applicable, I'm resubmitting it. I've heavily polished and simplified it.

It simply adds a new parameter to the /users POST api call that allows to trigger identical behaviour to adding users from the admin panel: Setting the newly signed up user as already confirmed and with an expired password, resulting in a "You have been added to Gitlab, click here to set a password" email.

As a bonus, I've also fixed two tests for that API call that were succeeding for the wrong reasons.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,7 @@
 v 7.5.0
   - API: Add support for Hipchat (Kevin Houdebert)
   - Add time zone configuration on gitlab.yml (Sullivan Senechal)
+  - Add force_random_password option to user POST api
   - Fix LDAP authentication for Git HTTP access
   - Fix LDAP config lookup for provider 'ldap'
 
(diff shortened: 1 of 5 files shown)


dblessing:

PR #7883, opened 2014-09-27
[WIP] Factor blob and blame [failure unrelated]

cirosantilli:
Philosophy: blame is just a type of view for a blob. There is even a toggle button to go between both.

This PR factors:
- the filename holder on blame and blob views
- controller logic for blame which is identical for blame by moving `blame` into the blob controller

It makes the following UI modifications which increases uniformity between blobs and blames:
- show the branch / path breadcrumb and last commit to the top of the blame
- remove the "blame view" title, since this is already obvious from the view itself, and from the "normal view" toggle button. Same rationale why the show view does not have a "Show view" title.

New:

screenshot from 2014-09-28 00 18 17 blob blame new

Old:

screenshot from 2014-09-28 00 18 32 blob blame old

We now have more functionality and uniformity with less lines of code.

TODO: the only implementation detail I'm not satisfied with is the route. Can anyone propose a shorter, DRYer way of writing it, possibly as a `member` of the `:blob` resource?

### DIFF
--- a/app/controllers/projects/blob_controller.rb
+++ b/app/controllers/projects/blob_controller.rb
@@ -39,6 +39,11 @@ class Projects::BlobController < Projects::ApplicationController
     render layout: false
   end
 
+  def blame
+    @blame = Gitlab::Git::Blame.new(project.repository, @commit.id, @path)
+    render :show
+  end
+
   private
 
   def blob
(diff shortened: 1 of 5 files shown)


jvanbaarsen:
@cirosantilli I'll close this for now, once you have time to work on this, just ping me and I'll re-open this PR.

PR #7905, opened 2014-09-29
Change always passing visible false tests.

cirosantilli:
Many usages of `visible: false` are wrong since `visible: false` matches _both_ visible and invisible elements: http://rubydoc.info/github/jnicklas/capybara/master/Capybara/Node/Finders#all-instance_method

The correct way to check that an element is present but not visible is:

```
expect(find(...)).not_to be_visible
```

The two "I can't preview without text" scenarios are not true anymore: now you can preview without text, so I removed then. They only passed because of the misuse. Screenshot:

screenshot from 2014-09-30 15 56 00 js not preview visible without text

`'I should see add a diff comment button'` is simply wrong: it should be `visible: true` instead of `visible: false`.

### DIFF
--- a/features/project/commits/comments.feature
+++ b/features/project/commits/comments.feature
@@ -14,11 +14,6 @@ Feature: Project Commits Comments
     Then I should not see the cancel comment button
 
   @javascript
-  Scenario: I can't preview without text
-    Given I haven't written any comment text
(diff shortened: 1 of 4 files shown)


cirosantilli:
@jvanbaarsen updated.

PR #7912, opened 2014-09-30
Convert many JavaScript # links to buttons [failure unrelated]

cirosantilli:
Same rationale as: https://github.com/gitlabhq/gitlabhq/pull/7863

Found with `git grep "'#'"`.

Refactored for every place except projects.

Only touched buttons that look like buttons, not links. Those should work with `.btn.btn-link` but for some reason didn't.

There were several dropdown ones factored with:

```
find . -iname '*.haml' | xargs perl -lapi -e "s/%a.dropdown-toggle.btn{href: '#', \"data-toggle\" => \"dropdown\"}/%button.dropdown-toggle.btn{type: 'button', 'data-toggle' => 'dropdown'}/"
```

We should really DRY up those dropboxes before they start diverging.

### DIFF
--- a/app/views/admin/projects/index.html.haml
+++ b/app/views/admin/projects/index.html.haml
@@ -44,7 +44,7 @@
         Projects (#{@projects.total_count})
         .panel-head-actions
           .dropdown.inline
-            %a.dropdown-toggle.btn{href: '#', "data-toggle" => "dropdown"}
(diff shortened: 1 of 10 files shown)


dblessing:
@cirosantilli It looks like this can be pulled into 7.8 if you can please rebase.

cirosantilli:
@dblessing failure does not happen locally

PR #7918, opened 2014-09-30
improve new project options

bbodenmiller:
Before:
image

After:
image


jvanbaarsen:
@bbodenmiller I would rather have the "Existing Repo?" copy A bit better then add an entire new block. When you have an existing folder it does not mean you still have to add a README and such. I'm closing this PR for now. But still ping @randx so he can decide if he wants to keep it closed or maybe re-opens it.

@randx If you still want this in, can you please re-open and merge this?

PR #7924, opened 2014-09-30
Replace defined? in views with local_assigns

cirosantilli:
`local_assigns` is the right way to do it: http://stackoverflow.com/questions/2060561/optional-local-variables-in-rails-partial-templates-how-do-i-get-out-of-the-de

One occurrence, `defined?(snippet[:start_line])`, is a bug, since `a = {}; defined? a[:b]` returns `"method"`, not `nil`.

### DIFF
--- a/app/views/layouts/_head.html.haml
+++ b/app/views/layouts/_head.html.haml
@@ -10,7 +10,7 @@
   %meta{content: "GitLab Community Edition", name: "description"}
 
   %title
-    = "#{title} | " if defined?(title)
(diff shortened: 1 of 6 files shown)


dblessing:
@cirosantilli I don't see any documentation references to `local_assign` anymore. Further, I see http://stackoverflow.com/questions/851258/should-the-extensive-use-of-local-assigns-has-key-in-partial-be-considered-a-co that suggests it's a poorly documented feature at best and questions its use. Is using `defined?` currently causing issues? If not, I'm not sure there's compelling reasons to change at this time.

PR #7925, opened 2014-09-30
Remove or prepend underscore _ to unused method arguments

cirosantilli:
- remove `up_dir_path tree` which can be removed
- add underscore Devise callbacks to silence Hound in those cases where it cannot be omitted as mentioned at: https://github.com/gitlabhq/gitlabhq/pull/7858#discussion-diff-18063612


randx:
Thanks

PR #7928, opened 2014-10-01
Remove empty query params from many filters [failure unrelated]

cirosantilli:
Before this PR, clicking on filters like `sort` at the issues index would give a URL with many empty parameters like:

```
assignee_id=&label_name=&milestone_id=&scope=all&sort=newest&state=opened
```

Now, `assignee_id=&label_name=&milestone_id=` only show if explicitly set, so the URL would be:

```
scope=all&sort=oldest&state=opened
```

The change should affect issues and merge request in indexes, group and global searches.

This works because we were using the built-in helper `url_for`, which removes keys which have nil values from params, while before params was being built with `to_params` which does not.

`scope` and `state` should also be removed if not set, but implementation there is more delicate because those params are always being set from controller and used in the logic as in https://github.com/gitlabhq/gitlabhq/blob/0f12e051c072521460fadfce2cec4e9768bc35ba/app/controllers/projects/issues_controller.rb#L131

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -305,4 +305,10 @@ module ApplicationHelper
       profile_key_path(key)
     end
   end
+
+  # Generate a path to the current URL with given url_for options.
+  # Existing whitelisted params are used as defaults if missing.
+  def merge_params_path(options, whitelist)
+    url_for(params.slice(*whitelist).merge(options))
+  end
 end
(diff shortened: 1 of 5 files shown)


cirosantilli:
@jvanbaarsen updated. Failure is unrelated and random: linked to the apostrophe encoding `'` on test data.

jvanbaarsen:
@cirosantilli Would it be possible for you to submit a patch that fixes this testing issue? That way we don't run into it all the time :) Can't merge this while its red :(

jvanbaarsen:
@cirosantilli I'm working on a patch for the testing issues.

jvanbaarsen:
@cirosantilli Can you rebase with master? My fix has merged

jvanbaarsen:

PR #7937, opened 2014-10-02
[WIP] Make issue assigned filters more uniform across group and project [Can be updated]

cirosantilli:
This PR makes the `Assigned to me / created by me / everyone's` issue filter more uniform between project, dashboard and group issues.

The only UI change is that for the dashboard and group, the order changed from:
- `assigned to me`
- `created by me`
- `everyone's`

to:
- `everyone's`
- `assigned to me`
- `created by me`

which matches the order under projects. Those two should have the same order: it makes the UI more uniform.

Note that the default values for those attributes is different between views and has not been changed: for dashboard it is `assigned to me`, for groups depends on signed in or not, and for project issues it is always everyone's.

The greater uniformity also allowed a large factorization. 

screenshot from 2014-10-02 11 11 14 factor filter state assigned

Implementation notes:
- removed the  `.scope-filter` class which is not used anywhere
- removed the unused `entity` argument to facilitate polymorphism

### DIFF
--- a/app/finders/issuable_finder.rb
+++ b/app/finders/issuable_finder.rb
@@ -6,7 +6,7 @@
 #   klass - actual class like Issue or MergeRequest
 #   current_user - which user use
 #   params:
-#     scope: 'created-by-me' or 'assigned-to-me' or 'all'
+#     scope: 'authored' or 'assigned-to-me' or 'all'
 #     state: 'open' or 'closed' or 'all'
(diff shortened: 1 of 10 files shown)


dblessing:
@cirosantilli Can you please fix the hound warnings and then we'll see if tests pass.

cirosantilli:
I'm a bit busy for this one now.

PR #7951, opened 2014-10-03
Make blob new and edit file editors more uniform

cirosantilli:
DRY them up and make them uniform. UI changes:
- remove cancel button from file edit header: there is already a Cancel button the bottom of the page, which is the standard place for cancel buttons
- add branch to new file edit. Already present on the edit, so let's make it uniform.

After edit: no cancel:

screenshot from 2014-10-03 14 17 30 factor editor after

Before edit:

screenshot from 2014-10-03 14 25 18 blob edit before

After new (has branch on header):

screenshot from 2014-10-03 14 24 44 blob new after

Before new:

screenshot from 2014-10-03 14 25 29 blob new before

### DIFF
--- a/app/views/projects/_blob_editor.html.haml
+++ b/app/views/projects/_blob_editor.html.haml
@@ -1,0 +1,15 @@
+.file-holder.file
+  .file-title
+    %i.icon-file
+    %span.file_name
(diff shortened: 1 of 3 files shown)


randx:
Thanks

PR #7977, opened 2014-10-06
Impossible to show a blob named `diff` [failure unrelated]

cirosantilli:
Before this commit, it was impossible to see a blob whose filename ends in `diff` because GET `path/to/diff` routed to `blob#diff`.

Currently live on GitLab: https://gitlab.com/cirosantilli/test/blob/master/diff

You will get redirected to https://gitlab.com/cirosantilli/test/tree/master/ where you can see the the file named `diff` does in fact exist!

This is an ugly fix that replaces the  GET with a POST to resolve the ambiguity but for something that does not change system state, but it is the simplest DRY strategy I could find, and the same strategy is currently used for blob edit preview at: https://github.com/gitlabhq/gitlabhq/blob/30c447ed2f35a0d1f4eb0f6200aa9ab952c46768/config/routes.rb#L201

To solve the problem beautifully, we must instead use a route like `diff_blob/:id`, and to do that nicely someone must answer: http://stackoverflow.com/questions/26223317/how-to-put-a-member-of-a-routing-resource-outside-of-its-parent-namespace

### DIFF
--- a/app/assets/javascripts/diff.js.coffee
+++ b/app/assets/javascripts/diff.js.coffee
@@ -31,7 +31,7 @@ class @Diff
         offset: offset
         unfold: unfold
 
-      $.get(link, params, (response) =>
+      $.post(link, params, (response) =>
         target.parent().replaceWith(response)
       )
     )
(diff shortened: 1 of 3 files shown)


cirosantilli:
@jvanbaarsen updated. I cannot reproduce the failure locally.

jvanbaarsen:
@cirosantilli I've retriggered the build, hopefully that helps :)

jvanbaarsen:

PR #8015, opened 2014-10-11
Add timezone configuration to gitlab.yml

soullivaneuh:
This pull request refer to this suggestion: http://feedback.gitlab.com/forums/176466-general/suggestions/5890076-add-timezone-configuration-to-gitlab-yml

It's now possible to set time zone form gitlab.yml configuration file by adding a little time_zone initializer.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,6 @@
+v 7.5.0
+  - Add time zone configuration on gitlab.yml (Sullivan Senechal)
+
 v 7.4.0
   - Refactored membership logic
   - Improve error reporting on users API (Julien Bianchi)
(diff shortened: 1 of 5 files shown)


JonathonReinhart:
When searching for "gitlab timezone" it's very easy to find all of these previously-open issues, but it's difficult to find the new documentation telling you how you're *supposed* to do it in the newest versions.  

Would someone kindly add a link to the documentation for how to set the timezone in both the cookbook and omnibus versions?

For example, this change involves adding it to gitlab.yml, but shouldn't omnibus users be adding it to `/etc/gitlab/gitlab.rb`? Because `/opt/gitlab/embedded/service/gitlab-rails/config/gitlab.yml` indicates that `/etc/gitlab/gitlab.rb` is where changes should be made.

snobear:
@JonathonReinhart I've been unable to find any docs on this as well, but figured it out for Omnibus install.

**/etc/gitlab/gitlab.rb**

```
gitlab_rails['time_zone'] = 'America/New_York'
```

List of timezones [here](http://en.wikipedia.org/wiki/List_of_tz_database_time_zones).

**Reconfigure and restart**

```
sudo gitlab-ctl reconfigure
sudo gitlab-ctl restart
```

PR #8040, opened 2014-10-14
Use full path from root of project.

cmluciano:


dblessing:
The best thing is that this comment tries to use markdown format in a config file. @drag00n Can you please update the text to remove the markdown syntax? At that point I think it's great to have the full relative path like you've done.

drag00n:
@dblessing How does the latest commit look?

dblessing:
:+1:

dblessing:
Can you please squash the commits then I'll merge.

drag00n:
Squashed, thanks for taking a look at this @dblessing

PR #8095, opened 2014-10-19
Only redirect project path.git to path on #show

cirosantilli:
- POST path.git should not be allowed
- path.git/some/path currently already redirects to Grack

Makes the all project controllers faster by removing that check from all actions to only project#show.

Also added some tests.

This works because before hooks on base classes are run BEFORE those of parents, so the redirect is done before the `:project` before filter of `ApplicationController`.


jvanbaarsen:
@cirosantilli Can you please make this mergeable?

jvanbaarsen:

PR #8098, opened 2014-10-19
Only ignore Status changed note emails if system, not if made by users.

cirosantilli:
Currently, if you make an issue comment that starts with the magic string:

```
_Status changed to closed_
```

no notifications are generated.

After this PR they are, using the same mechanism that fixed this problem for `_mentioned in` notes, by first doing a `note.system` check.

### DIFF
--- a/app/services/notification_service.rb
+++ b/app/services/notification_service.rb
@@ -118,8 +118,13 @@ class NotificationService
     return true unless note.noteable_type.present?
 
     # ignore gitlab service messages
-    return true if note.note =~ /\A_Status changed to closed_/
-    return true if note.note =~ /\A_mentioned in / && note.system == true
+    if note.system
+      note_content = note.note
+      if note_content.start_with?('_Status changed to closed_') ||
+         note_content.start_with?('_mentioned in ')
+        return true
+      end
+    end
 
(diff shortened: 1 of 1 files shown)


dblessing:
@randx What do you think about the indentation style here? If you approve I'll request a rebase.

app/services/notification_service.rb
```diff
@@ -118,8 +118,13 @@ def new_note(note)
     return true unless note.noteable_type.present?
 
     # ignore gitlab service messages
-    return true if note.note =~ /\A_Status changed to closed_/
-    return true if note.note =~ /\A_mentioned in / && note.system == true
+    if note.system
+      note_content = note.note
+      if note_content.start_with?('_Status changed to closed_') ||
+         note_content.start_with?('_mentioned in ')
```
randx: Indentation should be 2 spaces I think. So for current case

```
if foo && 
  bar
  puts 'echo'
end
```

randx: I think it should be everytime 2 spaces everywhere except this case: 

```
var = VeryLongClass.method(foo: bar, 
                           bar: foo) 
```

jvanbaarsen:

PR #8086, opened 2014-10-18
Create merge request when editing on web UI v3 [can be updated]

cirosantilli:
Fixes: ACCEPTING MR at: http://feedback.gitlab.com/forums/176466-general/suggestions/5350102-option-to-make-a-merge-request-when-editing-via-we

Supersedes the original request: https://github.com/gitlabhq/gitlabhq/pull/7266

I proposed custom branch names to GitHub at: https://github.com/isaacs/github/issues/228 and they've half implemented it at: https://github.com/blog/1945-quick-pull-requests 

Dependencies:
- [x] https://github.com/gitlabhq/gitlabhq/pull/8053/files (fixes the only CI failure)
- [ ] https://github.com/gitlabhq/gitlabhq/pull/8257

Dependencies already included here which were extracted from this PR and should be evaluated separately and merged first:
- [ ] https://github.com/gitlabhq/gitlabhq/pull/8128 
- [ ] https://github.com/gitlabhq/gitlabhq/pull/8131
- [ ] https://github.com/gitlabhq/gitlabhq/pull/8183

TODO on future PRs:
- Allow to create new branch without creating a merge request (add new checkbox, implied by create mr)
- When already on fork, add a "Create merge request back to origin checkbox"

TODO just before merging this after having evaluated:
- set the gitlab-shell URL back to the original one
## Blob edit page UI changes

If user has push permission on the current branch, when the page loads it looks like:

screenshot from 2014-11-02 17 12 20 initial

If user clicks on "Create merge request", he now sees:
- the "On my fork" checkbox which shows where the merge request source will be
- the source branch indicator `master` becomes an input so he can choose a custom branch name
- the source repository changes from `gitlab-org` to `root`

screenshot from 2014-11-02 17 13 10 click create mr

By default, branch name is a free branch name on both repositories. When user toggles "On my fork", and he has not yet modified it, it toggles the branch name to a free branch name on given repo:

screenshot from 2014-11-02 17 32 43 on my fork

If user edits the branch name to a custom name like `custom-name`, the branch name toggle stops so as not to lose the user's input. So after editing the branch name and toggle "On my fork", the branch name stays at `custom-name`:

screenshot from 2014-11-02 17 34 40 custom name

User does not have push permission, he must fork. "Create merge request" and "On my fork" are checked and disabled:

screenshot from 2014-10-15 21 23 22 user does not have push permission on current branch

If user is already on his fork, he does not see "On my fork": the merge request will necessarily be generated between the current repo and itself:

[...]

PR #8128, opened 2014-10-22
Fix gitlab shell test preparation reset method

cirosantilli:
Bug introduced by self at: https://github.com/gitlabhq/gitlabhq/pull/7823

Before this PR, the test gitlab shell under `tmp/tests/gitlab-shell` will not correctly fetch new commits if necessary when it exists already (the cache was introduced on the PR that causes this problem: before it would always get removed and cloned).

This happens because the `shell.rake` `sh reset` line I wrote was completely wrong because:
1. bad understanding of `git reset --hard`: if there is no reference, `reset --hard $()` will do just `git reset --hard`, which is valid, and does not activate the `||`
2. bad understanding of precedence: `true || echo a && echo b` translates to `(true || echo a) && echo b`, so the `fetch` part was never done!

Now it works, but is a monster expression. We could split this up into multiple system calls, but that would break the current convention of using `sh`.

This was extracted from: https://github.com/gitlabhq/gitlabhq/pull/8086

### DIFF
--- a/lib/tasks/gitlab/shell.rake
+++ b/lib/tasks/gitlab/shell.rake
@@ -24,8 +24,10 @@ namespace :gitlab do
       Dir.chdir(target_dir) do
         # First try to checkout without fetching
         # to avoid stalling tests if the Internet is down.
-        reset = "git reset --hard $(git describe #{args.tag} || git describe origin/#{args.tag})"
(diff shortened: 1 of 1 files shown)


jvanbaarsen:
@cirosantilli Can you please make this mergeable?

jvanbaarsen:

Issue #4137, opened 2013-05-30
Importing existing repositories directly from the repositories path requires clarification.


tnarik:
Related to #3693, #3297, #2932, #3636, #3926 and https://github.com/gitlabhq/gitlabhq/wiki/Import-existing-repositories-into-GitLab.

The way I see it:

1. Bare repositories copied to /home/git/repositories and imported with 5.x are not copied but just created inside the 'admin' namespace
  For instance: ```/home/git/repositories/repo.git``` will be left there and a new and empty ```/home/git/repositories/admin/repo.git``` will be created.
2. Instead, if #2932 is used to patch the system, the repository will be used as is.
  For instance: ```/home/git/repositories/repo.git``` will be added to GitLab and files will be available.
3. ```rake gitlab:import:repos``` (from 5.2) only works from /home/git/repositories, not any subfolder, so without any changes, it will create empty repos under the 'admin' namespace.
Then the existing repositories can be moved on top of the new ones, replacing for instance the empty ```/home/git/repositories/admin/repo.git``` with the contents of ```/home/git/repositories/repo.git```, which will make files available.

It would be great if the wiki is updated to reflect the current solution (which as far as I know is as described in 3.) and additionally the way forward is described (commit 5f9d78fdff746a0b50313cf415acac809569a186 is meant to allow importing of repositories inside namespaces [and it works in spite of displaying an error because it cannot create the 'admin' group], but it doesn't seem to take care of the global namespace, for what I saw).

jvanbaarsen:
We are currently in the process of closing down the issue tracker on GitHub, to prevent duplication with the GitLab.com issue tracker. Since this is an older issue I'll be closing this for now. If you think this is still an issue I encourage you to open it on the [GitLab.com issue tracker](https://gitlab.com/gitlab-org/gitlab-ce/issues).

PR #8140, opened 2014-10-22
Only search gitlab:import:repos on repos on subdir [failure unrelated][can be updated]

cirosantilli:
of type `repos_path/group/repo.git`.

Before this, any other location like `repos_path/repo.git`
or `repos_path/group/XXX/repo.git` does generate new projects, but
these projects are empty since no copy is done.

Fix https://github.com/gitlabhq/gitlabhq/issues/4137

This behavior probably exists to support the old global repositories, but since it is broken already, I propose we remove support for that once and for all here. Let users put their repositories on the regular repository structure only which is less confusing and easier to maintain.

Implementation notes: 
- `group_name == '.'` is not possible since glob does not expand to `.`, so that logic was removed.

### DIFF
--- a/doc/raketasks/import.md
+++ b/doc/raketasks/import.md
@@ -22,6 +22,8 @@ your repositories are located by looking at `config/gitlab.yml` under the `gitla
 $ cp -r /old/git/foo.git/ /home/git/repositories/new_group/
 ```
 
+The repository folder must end in `.git`.
+
 ### Run the command below depending on your type of installation:
 
 #### Omnibus Installation
--- a/lib/tasks/gitlab/import.rake
+++ b/lib/tasks/gitlab/import.rake
@@ -13,7 +13,7 @@ namespace :gitlab do
     task repos: :environment do
 
       git_base_path = Gitlab.config.gitlab_shell.repos_path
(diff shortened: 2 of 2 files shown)


jvanbaarsen:
Hm the weird thing is, one built is successful, and the other is not.. Can you rebase? Maybe it works after that :)

jvanbaarsen:

jantman:
Is this just being ignored now? I just spent half an hour going through the MANY wiki articles and google results trying to figure out why the repositories that I imported [per the docs|https://github.com/gitlabhq/gitlabhq/blob/master/lib/tasks/gitlab/import.rake] were empty, before I found this PR.

cirosantilli:
@jantman GitLab ignores PRs if we don't update them in 2 weeks after request. They took too long to review and I had lost interest when they requested update.

Feel free to pull this, make it mergeable and a new PR. If you make this mergeable, there is some chance that it will get merged.

I think it's not a very good GitLab policy, they should at least tag non updated PRs, which I do like `[can be updated]`.

Also I think that if something is worth being reviewed and merged, it is also worth being rebased.

:-)

PR #8149, opened 2014-10-23
Use send only where necessary

cirosantilli:
Prefer public_send if methods are public,
and method calls if the method is a constant symbol.

`public_send` mentioned in bbatsov: https://github.com/bbatsov/ruby-style-guide#prefer-public-send

Under `app` and `lib`:

```
find . -iname "*.rb" | xargs perl -lapi -e 's/\.send\b/.public_send/'
```

but some manual refactor was also done in addition.

Exceptions:
- `self.send` and just `send`. I moved `self.send` to just `send` for uniformity and to avoid hound barks. In those cases, the call can access private methods, so it does not matter.
- at `lib/redcarpet/render/gitlab_html` there is an `h.send` which causes problems is replaced... seems like we are using something we shouldn't.

### DIFF
--- a/app/finders/issuable_finder.rb
+++ b/app/finders/issuable_finder.rb
@@ -45,7 +45,7 @@ class IssuableFinder
 
     if project
       if project.public? || (current_user && current_user.can?(:read_project, project))
-        project.send(table_name)
+        project.public_send(table_name)
       else
         []
       end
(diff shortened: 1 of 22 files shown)


jvanbaarsen:
@cirosantilli I really would like you to change the symbol literals :-) See it like the boyscout rule, leave it a little better then you found it :) it really does make things more clear when scanning over the code to see `"some_string".to_sym` instead of `:"some_string"`

jvanbaarsen:

PR #8172, opened 2014-10-26
Stop blacklisting project path

cirosantilli:
Before this PR, all project paths at: https://github.com/gitlabhq/gitlabhq/blob/199029b842de7c9d52b5c95bdc1cc897da8e5560/lib/gitlab/blacklist.rb were forbidden.

This PR removes that restriction, keeping it only for namespaces where it is required.

The restriction is not necessary and just creates more exception cases that can annoy users.

The blacklist was used here because of the fear of conflicting an URLs like `/groupname/issues` for two possible uses:
1. all issues of the group
2. a project called `issues`

however case 1 is impossible since:
- all group URLs are prefixed by `/group` (e.g. `/group/groupname/issues`), except the base `/groupname` which redirects to `/group/groupname`
- currently URLs like `/groupname/issues` don't redirect to `/group/groupname/issues` and give 404 instead.
  
  This is sane and should not be undone in the future, or else we would have problems whenever we want to add a new path to the blacklist.

This is also the URL strategy used by GitHub.

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -98,7 +98,6 @@ class Project < ActiveRecord::Base
             format: { with: Gitlab::Regex.project_name_regex,
                       message: Gitlab::Regex.project_regex_message }
   validates :path, presence: true, length: { within: 0..255 },
-            exclusion: { in: Gitlab::Blacklist.path },
             format: { with: Gitlab::Regex.path_regex,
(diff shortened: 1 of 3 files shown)


jvanbaarsen:

Razer6:
Already removed.

PR #8180, opened 2014-10-27
Trigger merge request hook when source updated

kzaitsev:
**What does this MR do?**
Allow trigger merge request web hook when source project updated

**Are there points in the code the reviewer needs to double check?**
https://github.com/Bugagazavr/gitlabhq/blob/mr_update/app/models/project.rb#L424-L425

**Why was this MR needed?**
Allow detect source changes in MR for many external systems


randx:
@Bugagazavr can you send a PR with CHANGELOG entry for this change?

PR #8225, opened 2014-11-02
Link to create blob from blob show [Awaiting developper feedback][Can be updated]

cirosantilli:
After: add a `+` sign link to create a new blob. Clicking on it will create a new file in current directory.

screenshot from 2014-11-02 19 15 43 after

Before:

screenshot from 2014-11-02 19 15 36 before

This would make it analogous to many other show pages which also have a create new button, like tree show:

screenshot from 2014-11-02 19 15 52 tree

and wiki pages:

screenshot from 2014-11-02 19 16 14 wiki

If concept OK I'll implement.

### DIFF
--- a/app/views/projects/blob/_blob.html.haml
+++ b/app/views/projects/blob/_blob.html.haml
@@ -14,6 +14,11 @@
           = link_to truncate(title, length: 40), project_tree_path(@project, path)
       - else
         = link_to title, '#'
+    &nbsp;
+    &nbsp;
+    = link_to 'test', title: 'New file', id: 'new-file-link' do
+      %small
+        %i.fa.fa-plus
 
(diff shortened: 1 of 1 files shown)


jvanbaarsen:
ping @randx

randx:
It looks a bit wrong to me. This link triggers file creation and makes a lot of sense to have it inside directory but it confuses me when placed near file itself. Summary: I dont think this is good from UI perspective

cirosantilli:
@randx In Vim I do `:e` and create a new file in the same directory. But maybe not all users are Vimmers.

randx:
@cirosantilli But in file explorer you go to the directory to create file in it. And I see GitLab file browsing more like this

PR #8243, opened 2014-11-04
Don't use File.join to join id branch path pairs as it is system dependent

cirosantilli:
An id is the terminology for a branch path pair like: 

```
master/path/to/file.md
```

It is not semantically correct to use File.join in those cases since that uses
the separator of the current system, but the GitLab frontend is
system agnostic, e.g., should in theory work in Windows where the
path separator is \ without front-end changes.

### DIFF
--- a/app/controllers/projects/blob_controller.rb
+++ b/app/controllers/projects/blob_controller.rb
@@ -47,7 +47,8 @@ class Projects::BlobController < Projects::ApplicationController
     if @blob
       @blob
     elsif tree.entries.any?
-      redirect_to project_tree_path(@project, File.join(@ref, @path)) and return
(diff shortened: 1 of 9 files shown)


jvanbaarsen:
Codewise this looks ok. But maybe @randx can take an extra look before merging

randx:
The chances GitLab will work on windows in next 5 years are ~ 0.01%. and by this time this fix will make no sense. I think this PR adds additional complexity but has no benefits for GitLab in close feature

randx:
Still thank you for working on improvements to GitLab.

cirosantilli:
@randx I understand. Joking now: but this goes against: https://twitter.com/gitlab/status/478990520505888769 >:)

PR #8325, opened 2014-11-16
Move issue title to the top of the page [Awaiting developper feedback]

cirosantilli:
Every time I look at a GitLab issue I feel my eyes wondering for 2 seconds to find the title.

Titles should be one of the first things we see on pages.

After: remove "Issue" since already on the tab, put the title there, de-emphasize the id with a gray color:

screenshot from 2014-11-16 18 51 02 after

Before:

screenshot from 2014-11-16 18 53 24 before

It gets even worse if there are headers in the description, since even an `h2` is larger than the title:

screenshot from 2014-11-16 18 54 08 before with header

This is also closer to how GitHub looks.

If you like this, it should also be done for merge requests and milestones simultaneously.

### DIFF
--- a/app/assets/stylesheets/generic/issue_box.scss
+++ b/app/assets/stylesheets/generic/issue_box.scss
@@ -93,7 +93,7 @@
   }
 
   .description {
-    padding: 0 15px 10px 15px;
+    padding: 10px 15px;
 
     code {
       white-space: pre-wrap;
--- a/app/assets/stylesheets/sections/issues.scss
+++ b/app/assets/stylesheets/sections/issues.scss
@@ -152,3 +152,11 @@ form.edit-issue {
     }
   }
 }
+
+body[data-page="projects:issues:show"] {
+  .page-title {
+    .id {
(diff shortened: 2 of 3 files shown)


jvanbaarsen:
@randx Can you give your feedback on this?

randx:
Does your problem applies to new UI?

cirosantilli:
@randx it's less bad now that we have less header clutter, but the same principle still applies: title should come on top.

randx:
I dont agree with you. I think issue title and description should be together. The thing we can do is actually reduce font size for headers in issue description so it doesn't look bigger than issue title

cirosantilli:
@randx it is subjective, but both GitHub and JIRA separate title from description with metadata.

randx:
You are right. It is subjective :)

jvanbaarsen:
@randx Does that mean @cirosantilli can continue working on this? Or will you not accept this PR?

randx:
No it means that it is subjective from both sides. So we will leave it as is for now

PR #8326, opened 2014-11-16
Factor your changes have been successful message [failure unrelated]

cirosantilli:
The message was repeated 3 times, surrounding logic 2 times.

### DIFF
--- a/app/controllers/projects/application_controller.rb
+++ b/app/controllers/projects/application_controller.rb
@@ -29,4 +29,20 @@ class Projects::ApplicationController < ApplicationController
       redirect_to project_tree_path(@project, @ref), notice: "This action is not allowed unless you are on top of a branch"
     end
   end
+
+  protected
+
+  def changes_successful_action(result, redirect_path)
+    if result[:status] == :success
+      flash[:notice] = changes_successful_action_message
+      redirect_to redirect_path
+    else
+      flash[:alert] = result[:message]
+      render :show
+    end
+  end
+
+  def changes_successful_action_message
+    'Your changes have been successfully committed'
+  end
 end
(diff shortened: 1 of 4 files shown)


app/controllers/projects/application_controller.rb
```diff
@@ -29,4 +29,20 @@ def require_branch_head
...
+      flash[:notice] = changes_successful_action_message
+      redirect_to redirect_path
+    else
+      flash[:alert] = result[:message]
+      render :show
+    end
+  end
+
+  def changes_successful_action_message
+    'Your changes have been successfully committed'
```
jvanbaarsen: What about moving this text to an I18n file?

jvanbaarsen:
@cirosantilli I don't really like this change, since this feel like DRY'ing up the code a little to much. Some logic is ok to be present multiple times if it makes the code more clear. I believe this is the case here. 

I will not close this right away, but first discuss this change with the rest of the core team.

randx:
@cirosantilli thank you for all refactoring and cleanup you do for GitLab. 

But for current case I think this is DRY'ing in wrong direction
Better move blob new/edit/destroy to one controller  and have common method here for this actions. Otherway we have all magic happens in ApplicationController which is not nice

jvanbaarsen:
@cirosantilli Im closing this MR for now. Thanks for all your effort so far though :heart:

cirosantilli:
@randx that refactor is necessary but non trivial because of the routes, and this moves us closer to it. It will be harder to do later if things start diverging.

randx:
@cirosantilli agree. Do you want to make the refactoring of blob actions(new/edit/destroy) to one controller

cirosantilli:
@randx sorry, no time.

PR #8327, opened 2014-11-16
Don't render old data project / group names tags [can be updated]

cirosantilli:
which could be added to the database before the current column
validation was in place.

Before this PR, sanitize would allow good tags to be rendered
and remove bad tags like script.

After this, any tags will be HTML escaped instead of sanitized,
so both bad and good tags will appear escaped.

Rationale: people who entered data earlier should not be able to
do things that newer users can't. When we forbid something from being done,
we should warn users who did it to prepare to migrate.

This continues the discussions at: https://github.com/gitlabhq/gitlabhq/pull/8107#issuecomment-61004035

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -147,11 +147,6 @@ module ApplicationHelper
     image_tag(image_path("authbuttons/#{file_name}"), alt: "Sign in with #{provider.to_s.titleize}")
   end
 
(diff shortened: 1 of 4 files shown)


jvanbaarsen:
@cirosantilli Im not really sure if I understand what this PR does, can you try and explain in a bit more detail?

jvanbaarsen:

PR #8329, opened 2014-11-16
Factor gfm escape_once into a method

cirosantilli:
The pair appears 7 times. Let's DRY that up.


jvanbaarsen:
I think the old code was a bit easier to read (you know exaclty what was happening). I'll close this PR for now. Still much thanks for the effort!

PR #8338, opened 2014-11-17
Remove unneeded || @commit.id [can be updated]

cirosantilli:
I think this is not necessary because the commit is always taken from the id: https://github.com/gitlabhq/gitlabhq/blob/e37461979a5eb890975125134e7717c470c76a5f/lib/extracts_path.rb#L105

unless `@options[:extended_sha1]` is present, which only happens for the network, and never for trees.

### DIFF
--- a/app/views/projects/tree/_blob_item.html.haml
+++ b/app/views/projects/tree/_blob_item.html.haml
@@ -2,7 +2,7 @@
   %td.tree-item-file-name
     = tree_icon(type)
     %span.str-truncated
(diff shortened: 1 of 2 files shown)


jvanbaarsen:
@randx Do you know why the `@commit` was used?

randx:
@jvanbaarsen I have no idea :) I think it was a case earlier when `@id` was nil sometimes

jvanbaarsen:
@cirosantilli Can you please make this mergeable again?

PR #8343, opened 2014-11-18
Make web hooks retry on non-200 status codes from remote server.

### LINKED ISSUE #5768: Web/System hooks not retried if remote server returns non-200 error code.
It seems that executing Web and System hooks ignores the return code of the remote server. As a result a hook calls with return code 4xx or 5xx from the remote server are considered successful and not retried.

See https://github.com/gitlabhq/gitlabhq/blob/master/app/models/web_hook.rb#L31.

Perhaps if the return code should be checked and exception raised on non-200 return code?

I can submit a PR, if you think this is the right way to go, @randx.

bladealslayer:
Web hooks are previously handled in a fire-and-forget manner. If the remote server returns a non-2XX status code, the hook is never retried.

With this change, an exception is raised on non-2XX status code and this the hook is scheduled for retry.

Fixes #5768.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -62,6 +62,7 @@ v 7.8.0
   - 
   - 
   - Added support for firing system hooks on group create/destroy and adding/removing users to group (Boyan Tabakov)
+  - Make web hooks raise exception on non-200 status codes so that sidekiq retries failed requests (Boyan Tabakov)
 
 v 7.7.2
   - Update GitLab Shell to version 2.4.2 that fixes a bug when developers can push to protected branch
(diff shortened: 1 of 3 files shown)


jvanbaarsen:
With the current implementation changes are that webhooks are retried endlesly. I would rather see a way that a webhook is only retried N times and then fails gracefully. What do you think?

bladealslayer:
@jvanbaarsen my understanding is that sidekiq does exponential backoff with a retry limit, so there shouldn't be any endless retrying.

jvanbaarsen:
@randx What do you think about having Sidekiq go in an retry mode? I think the retry system from sidekiq is mostly used if something weird went wrong. And not in known cases.

randx:
I dont like it. There were a lot of cases when URL for webhook never returned an 2xx. It causes enormous amount of useless retries and huge exception trash in sidekiq logs

jvanbaarsen:
@bladealslayer Thank you for your effort, but we will not merge this as it is know.

PR #8356, opened 2014-11-19
Web edit and HTTPS push work if multiple Rubies installed v3 [can be updated]

cirosantilli:
Would fail because Git modifies the hooks PATH, so for example
the at /usr/bin/ruby could be called instead of the correct one
at /usr/local/bin/ruby. Same for RVM.

Updates the reversed https://github.com/gitlabhq/gitlabhq/pull/8340 , same goals.

Requires: https://github.com/gitlabhq/gitlab-shell/pull/199/files

The new approach is to generate a file to be sourced with the PATH for gitlab shell on gitlab's startup, much like `.gitlab_shell_secret`.

Old approach with env vars failed because SSH does not pass through GitLab so the vars were not being set.

Tests were removed since they were breaking something on the GitLab internal server. I can add again if desired.


bb-generation:
I just tried to apply this pull request (and https://github.com/gitlabhq/gitlab-shell/pull/199/files for gitlab-shell) and it works perfectly.

My way of trying was:
* stop gitlab
* create patch for both pull requests (clone repository, ```git format-patch <sha1>```, change to installation, ```git apply <patch-file> --exclude <CHANGELOG VERSION|GITLAB_SHELL_VERSION>```)
* start gitlab again
* push via ssh to a repository
* edit a file with the web viewer

And everything I tested worked as it should.

jangrewe:
It would be great if this could be merged! I'm currently unable to accept any MRs because i need to have ruby-1.9.3 installed as the system ruby.

jangrewe:
I have just (manually) patched this into my running 7.7.2 instance, and it works perfectly fine! 
I only needed to also run the 'gitlab:shell:install' rake task so that the 'gitlab-shell/hooks/.env' file was created - and that should be done before creating the hook scripts, otherwise they're overwritten by the rake task.

Please please please merge this for 7.8.0!

costela:
Is there a reason why you can't simply use the contents of the GITLAB_PATH_OUTSIDE_HOOK env var that is already set when calling the hooks? This way you can have a simple wrapper (symlinked to all hooks), that does:
```bash
#!/bin/sh
if [ ! -z "$GITLAB_PATH_OUTSIDE_HOOK" ]; then
  export PATH=$GITLAB_PATH_OUTSIDE_HOOK
fi
exec $0.rb "$@"
```
And you don't need the initializer or the hidden .env file.
Or am I missing something obvious?

cirosantilli:
@costela let's move this to: https://github.com/gitlabhq/gitlab-shell/pull/186 Not using env vars because SSH won't see them.

jvanbaarsen:
@cirosantilli Can you please make it mergeable again?

cirosantilli:
@jvanbaarsen sorry, but I don't have much time to touch GitLab now :-s I'll mark it as `[can be updated]` so maybe someone can do it. Cheers.

jvanbaarsen:
@cirosantilli Thanks for responding :)

PR #8395, opened 2014-11-25
Socket [::]:123 on Linux listens on IPv4 and IPv6

sbeh:
This will ensure nginx starts up without the following errors messages:
nginx: [emerg] bind() to [::]:443 failed (98: Address already in use)
nginx: [emerg] bind() to [::]:443 failed (98: Address already in use)
nginx: [emerg] bind() to [::]:443 failed (98: Address already in use)
nginx: [emerg] bind() to [::]:443 failed (98: Address already in use)
nginx: [emerg] bind() to [::]:443 failed (98: Address already in use)
nginx: [emerg] still could not bind()

Googling for them leads you to this site:
https://chrisjean.com/2014/02/10/fix-nginx-emerg-bind-to-80-failed-98-address-already-in-use/

### DIFF
--- a/lib/support/nginx/gitlab-ssl
+++ b/lib/support/nginx/gitlab-ssl
@@ -39,7 +39,7 @@ upstream gitlab {
 ## Redirects all HTTP traffic to the HTTPS host
 server {
   listen 0.0.0.0:80;
-  listen [::]:80 default_server;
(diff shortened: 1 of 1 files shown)


Razer6:
@maxlazio Does this also require a change in the omnibus package?

PR #8430, opened 2014-12-04
Simplify SSH fingerprint regexp extraction

cirosantilli:
`[\d\h]` is the same as `\h`


jvanbaarsen:
@cirosantilli Yeah semaphore seems to have passes :-) Marking this as ready for merge! Thanks :)

PR #8434, opened 2014-12-05
teamcity basic support

apazdnikov:
FR http://feedback.gitlab.com/forums/176466-general/suggestions/4139027-teamcity-integration

[This is a clone of https://github.com/gitlabhq/gitlabhq/pull/7570, because of a pull request bug.]

Simple Teamcity integration for MergeRequest build status and link on a build.

Algorithm: get status of the last build in Teamcity for concrete Teamcity.BuildConfiguration_ID and MergeRequest.SourceBranch and MergeRequest.Commits.Last
Have to create a build configuration Teamcity.BuildConfiguration_ID to trigger build on each push.
Have to create a survice user and give him viewer-access rights on Teamcity.BuildConfiguration_ID.

Algorithm is very simple but it is working for us since 2014/07/15 and everybody happy :-)

PR #8439, opened 2014-12-09
Teamcity integration using 8.1 rest api

glasspelican:
This allows gitlab to get build status from teamcity and to trigger builds on push

known issues
*cant read build Que
*triggering builds will only work for teamcity 8.1+

I am aware of the existing pull request #8434 

teamcity configuration
the teamcity server will need to have a user added for gitlab.
 the projects build configuration will need to use the git hash for the build number so that the build can be located from the api.
server for merge requests to build server must be configured to build all branches

### DIFF
--- a/app/models/project_services/teamcity_service.rb
+++ b/app/models/project_services/teamcity_service.rb
@@ -1,0 +1,116 @@
+class TeamcityService < CiService
+  include HTTParty
+
+  prop_accessor :teamcity_url, :build_type, :username, :password
+
+  validates :teamcity_url, presence: true,
+            format: { with: URI::regexp }, if: :activated?
+  validates :build_type, presence: true, if: :activated?
+  validates :username, presence: true,
+            if: ->(service) { service.password? }, if: :activated?
+  validates :password, presence: true,
+            if: ->(service) { service.username? }, if: :activated?
+
+  attr_accessor :response
+
+  after_save :compose_service_hook, if: :activated?
+
+  def compose_service_hook
+    hook = service_hook || build_service_hook
+    hook.save
+  end
+
+  def title
+    'JetBrains TeamCity CI'
+  end
+
+  def description
+    'A continuous integration and build server'
+  end
+
+  def help
+    'The build configuration in Teamcity must use the build format '\
+    'number %build.vcs.number% '\
+    'you will also want to configure monitoring of all branches so merge '\
+    'requests build, that setting is in the vsc root advanced settings.'
+  end
+
+  def to_param
+    'teamcity'
+  end
+
+  def fields
+    [
+      { type: 'text', name: 'teamcity_url',
+        placeholder: 'TeamCity root URL like https://teamcity.example.com' },
+      { type: 'text', name: 'build_type',
+        placeholder: 'Build configuration ID' },
+      { type: 'text', name: 'username',
+        placeholder: 'A user with permissions to trigger a manual build' },
+      { type: 'password', name: 'password' },
+    ]
+  end
+
+  def build_info(sha)
+    url = URI.parse("#{teamcity_url}/httpAuth/app/rest/builds/"\
+                    "branch:unspecified:any,number:#{sha}")
+    auth = {
+      username: username,
+      password: password,
+    }
+    @response = HTTParty.get("#{url}", verify: false, basic_auth: auth)
+  end
+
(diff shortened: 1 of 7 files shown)


app/models/project_services/teamcity_service.rb
```diff
@@ -0,0 +1,116 @@
...
+  def execute(data)
+    auth = {
+      username: username,
+      password: password,
+    }
+
+    branch = data[:ref]
+
+    self.class.post("#{teamcity_url}/httpAuth/app/rest/buildQueue",
+                    body: "<build branchName=\"#{branch}\">"\
```
opichals: When testing with 'TeamCity Enterprise 8.1.2 (build 29993)' the buildQueue POST data require the branchName attribute to contain just the branch name (just `xyzbranch` and not the full `ref` in the `refs/heads/xyzbranch`).

With the `refs/heads/xyzbranch` the build looks like it is working but always compiles the default branch (or master if not defined).

opichals:
I found out the TeamCity build is triggered for every commit to any branch.

What would you think about an option to only trigger the builds for open merge requests (could be configured in the service settings)?

app/models/project_services/teamcity_service.rb
opichals: Created a https://github.com/gitlabhq/gitlabhq/pull/8812 for this change.

glasspelican:
Ok looks like i didn't catch this because of how my ts is configured it didn't cause an issue.
I do like that idea
how about 
 - [ ] build all commits

opichals:
Or even better as a select box option which would name the desired trigger commits directly in the UI like:
Triggers: open merge requests commits | all commits

progmars:
I too waiting for this. It would be ideal to achieve the following:
- when Merge Request is created in GitLab, TeamCity receives a build trigger command with source and target branch names
- TeamCity clones git repo to get aclean local copy (or maybe just pulls)
- TeamCity merges source into target, builds the sources, rolls back the commit (because it is not pushable yet) and creates the results report.
- GitLab can pull the results from TeamCity and display them for Merge Request
- Project lead looks at the results, does some additional checks, if necessary, and approves the merge request, which then gets merged into master by GitLab.

For the time being maybe I can achieve something similar using GitLab webhooks, which will go to a custom gateway app which then translates it to a request for TeamCity build trigger API?

glasspelican:
>when Merge Request is created in GitLab, TeamCity receives a build trigger command with source and target branch names

right now it triggers builds for all commits not just merge requests
I have most of a fix for this in my dev env but have not had time to test it

>TeamCity clones git repo to get aclean local copy (or maybe just pulls)
>TeamCity merges source into target, builds the sources, rolls back the commit (because it is not pushable yet) and creates the results report

internal to teamcity, api gives me no control over this

>GitLab can pull the results from TeamCity and display them for Merge Request

Should be working

>For the time being maybe I can achieve something similar using GitLab webhooks, which will go to a custom gateway app which then translates it to a request for TeamCity build trigger API?

Is something not working the way you expect?

progmars:
Mostly everything is fine, but just that redundant build triggering for every commit.

I guess, I have two options then:
- wait for @glasspelican's fix
- add a webhook for "Merge Request events" and use that as a trigger. But then I'll have to implement some intermediate web service which translates from GitLab API format to TeamCity API.

The second option seems too complicated, so I will wait.

glasspelican:
I have been over-complicating this problem
looks like i can just edit the supported_events handle to include tag_push and merge_request
will be testing to make sure it works as expected, should have a pr in a few days

roend83:
Is there any documentation of how this is supposed to work and how you're supposed to configure TeamCity to work with it.  For example, does this plugin tell TeamCity to start a build or is TeamCity supposed to have VCS trigger?

ssube:
Has this feature been merged into Gitlab or is there another PR pending with fixes? I'm interested in setting this up for my Gitlab and TC servers.

glasspelican:
yes this was merged back in December 
but it looks like it may soon be outdated by jetbrains
https://twitter.com/gitlab/status/634719137168658433
https://youtrack.jetbrains.com/issue/TW-41219

progmars:
Will Jetbrains plugin support detection of GitLab merge requests to build only them instead of every commit? And how is it going, @glasspelican, with your fix - is it ready?

PR #8486, opened 2014-12-24
Gracefully shutdown services in Docker container

kfei:
The problem is `docker stop` only sends SIGTERM to the PID 1 inside the container, and the PID 1 (`/bin/sh -c ...`) does not take care of signals. Hence the services (e.g., postgresql, redis, sidekiq, etc) never have chances to graceful shutdown. Docker just kills the container after its 10 seconds timeout by default.

What this commit does:
- Add a wrapper as the entrypoint of Docker container. Which starts
  services through `runit`, reconfigure Gitlab by `gitlab-ctl` and
  gracefully shutdown all services when a SIGTERM is received.
- Create an `assets` directory for assets.
- Add `.dockerignore` file.

Now you'll see the following log messages after `docker stop`:

```
SIGTERM signal received, try to gracefully shutdown all services...
ok: down: logrotate: 1s, normally up
ok: down: nginx: 0s, normally up
ok: down: postgresql: 1s, normally up
ok: down: redis: 0s, normally up
ok: down: sidekiq: 0s, normally up
ok: down: unicorn: 0s, normally up
```

Signed-off-by: kfei kfei@kfei.net


ayufan:
It's generally a bad idea to use ENTRYPOINT instead of CMD. In that case it will make this command not working: `docker run -ti -e TERM=linux --rm --volumes-from gitlab_data ubuntu`.

kfei:
@ayufan Thanks for your reminder. I didn't notice that it conflicts with README.

I agree with you that ENTRYPOINT can sometimes confuse new Docker users. But it is required for dealing with signals. So I improved the wrapper to support the data volume container use case (kinda old-fashioned IMO) from README. Please have a look at that.

ayufan:
It is not entirely true. There are two variants of CMD (one with brackets
and not).

Kamil

On Thursday, January 8, 2015, LIN, KE-FEI <notifications@github.com> wrote:

> @ayufan <https://github.com/ayufan> Thanks for your reminder. I didn't
> notice that it conflicts with README.
>
> I agree with you that ENTRYPOINT can sometimes confuse new Docker users.
> But it is required for dealing with signals. So I improved the wrapper to
> support the data volume container use case (kinda old-fashioned IMO) from
> README. Please have a look at that.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/gitlabhq/gitlabhq/pull/8486#issuecomment-69147924>.
>


-- 
Kamil Trzciński

ayufan@ayufan.eu
www.ayufan.eu

kfei:
@ayufan You're right. I did not notice that CMD can also act as PID1. I've update my PR to use CMD instead. Thanks!

dosire:
Thanks everyone, please let me know when it's updated.

kfei:
@dosire It's updated. The commits on my branch are squashed. Please have a look at that.

dosire:
@kfei Can I run it based on the existing README documentation?

kfei:
@dosire Yes it's compatible with README now.

Razer6:
@kfei Can you rebase?

kfei:
@Razer6 Sure. Rebased!

dosire:
Thanks @kfei, I merged it, can you test once more to verify everything still works?

PR #8492, opened 2014-12-26
Remove GFM from titles

dblessing:
Remove markdown rendering from titles in a list. It doesn't make sense to render markdown with links to issues, etc in a list because it disjoints the link. For example 'This merge request fixes #1' will result in 'This merge request fixes' being a link to the merge request and '#1' is a link to issue #1. 

This change makes sense for merge request, issue and commit lists. For merge requests and issues markdown will continue to be rendered when the merge request/issue is viewed.

### DIFF
--- a/app/views/projects/commits/_commit.html.haml
+++ b/app/views/projects/commits/_commit.html.haml
@@ -3,7 +3,7 @@
     = link_to commit.short_id, project_commit_path(project, commit), class: "commit_short_id"
     &nbsp;
     %span.str-truncated
-      = link_to_gfm commit.title, project_commit_path(project, commit.id), class: "commit-row-message"
+      = link_to commit.title, project_commit_path(project, commit.id), class: "commit-row-message"
       - if commit.description?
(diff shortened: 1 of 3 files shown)


jvanbaarsen:
@dblessing I don't like that the emoji's are not showing in the titles anymore :( What do you think about creating a custom title parser that does not link to other instances, but still does the bold and emoji thing?

Razer6:
Emojis already use the html-pipeline to replace the markdown content with links to emojis as shown here: https://github.com/gitlabhq/gitlabhq/blob/master/lib/gitlab/markdown.rb#L77

The idea is to make different pipelines. Like issue/MR content, Titles, Dashboard content,..
The `Titles` pipeline would still have enabled the EmojiPipeline. The repository for the html-pipeline is here: https://gitlab.com/gitlab-org/html-pipeline-gitlab

dblessing:
I'm ok with looking at the pipeline solution. Let's close this for now.

PR #8495, opened 2014-12-26
Merge request error display

### LINKED ISSUE #8432: "You must select source and target branches" is always visible on page load
Whenever I visit the New Merge Request page, I see the error below.
bug
Seems to me it should only be visible when we try to create merge request by clicking the button without both source and target branches. As of now, it's always visible.
GitLab 7.5.2

dblessing:
Fixes #8432 

Currently, GitLab displays an error message whenever a new merge request is created. This is often confusing for users. There is another catch point that will show an error (see screenshot) if a user clicks "Compare branches" without specifying a source or target. If a source and target branch are not specified under other circumstances I think it's OK to quietly 'fail' without showing a message. I'm not sure if this is acceptable but I can't see any pitfalls so far. Please provide feedback.

Before:
screen shot 2014-12-26 at 3 28 35 pm

Existing and unchanged (if user clicks 'Compare branches' without specifying'):
screen shot 2014-12-26 at 3 25 03 pm

### DIFF
--- a/app/services/merge_requests/build_service.rb
+++ b/app/services/merge_requests/build_service.rb
@@ -13,7 +13,7 @@ module MergeRequests
       merge_request.target_branch ||= merge_request.target_project.default_branch
 
       unless merge_request.target_branch && merge_request.source_branch
-        return build_failed(merge_request, "You must select source and target branches")
+        return build_failed(merge_request, nil)
       end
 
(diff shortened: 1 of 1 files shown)


jvanbaarsen:
@dblessing Looks good :)

randx:
:+1:

PR #8510, opened 2014-12-29
add environment variable force to the documentation

koebi001:
This is necessary if you like to suppress the question for rebuilding the authorized_keys file. This is necessary for example if you would like to run the command unattended in a script.
[ci skip]

### DIFF
--- a/doc/raketasks/backup_restore.md
+++ b/doc/raketasks/backup_restore.md
@@ -155,6 +155,7 @@ Options:
 
 ```
 BACKUP=timestamp_of_backup (required if more than one backup exists)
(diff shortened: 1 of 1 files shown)


jvanbaarsen:

jvanbaarsen:
@randx I have been looking at the source code for backups, but I can't seem to find any information about the force env var. Can you determine if this is used and is needed to be documented?

randx:
@jvanbaarsen restore of backup execute another rake task for authorized_keys where `force` argument is available

jvanbaarsen:
@randx Ah ok :)

PR #8515, opened 2014-12-30
Key comment extraction for generic comments

cinderblock:
When adding public keys to my account, I love that I don't need to title them as it extracts part of my comment that I've attached to the text. However, I regularly add extra spaces to my comments for extra data (like creation date) and that extra data does not get passed on to the key's title.

Looking into it, it looks like the little helper js code just searches for any text that looks like an email address and adds that to the title. This seems wrong and too restrictive, IMHO.


jvanbaarsen:

cinderblock:
Rebased and re-submitted.

PR #8528, opened 2015-01-02
Raise group avatar filesize limit to 200kb, fixes #8527

### LINKED ISSUE #8527: Group avatar maximum file size discrepency
When uploading a group avatar on gitlab.com the maximum size listed against the select file button is 200kb, however when trying to save the group with an avatar of roughly 160kb I get the following error;

gitlab_avatar_size_bug

As you can see, the error message indicates a 100kb file size, yet you can still clearly see the conflicting 200kb limit shown next to the file selection button.

Razer6:
With GitLab 7.6, the file size limit for user avatars was raised from 100kb to 200kb. 

Since group avatars use the same view as user avatars there is now a conflict because UI says a limit of 200kb although the model validates against 100kb. This PR raises the file size limit for group avatars from 100kb to 200kb.

### DIFF
--- a/app/models/group.rb
+++ b/app/models/group.rb
@@ -21,7 +21,7 @@ class Group < Namespace
   has_many :users, through: :group_members
 
   validate :avatar_type, if: ->(user) { user.avatar_changed? }
-  validates :avatar, file_size: { maximum: 100.kilobytes.to_i }
+  validates :avatar, file_size: { maximum: 200.kilobytes.to_i }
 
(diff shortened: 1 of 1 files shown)


TeatroIO:
I've prepared a stage. <a href='http://razer6-fix-group-avatar.c68bfdc4cac4c32c41e1a922e73062ea.ttrcloud.com' target='_blank'>Click to open.</a>

PR #7394, opened 2014-07-29
Add Asana service

j0k3r:
Current state: WIP - Do not merge
- [x] working asana communication
- [x] ability to restrict to Branch (like the `master`)
#### What does this MR do?

When someone commit with a message including
- "blabla #123456", it will attach a new message to the task 123456 in Asana.
- "fix #456789", it will attach a new message and close (mark as done) the task 456789 in Asana

Also, I've added ability to render `service.help` as markdown. It's more easy to write the help message then.
#### Are there points in the code the reviewer needs to double check?
- `app/models/project_services/asana_service.rb`
- `app/views/projects/services/_form.html.haml`
#### Why was this MR needed?

We use Asana as our Issue tracking system. It will ease the relation between Asana and Gitlab by linking Asana tickets to git commit.
#### What are the relevant issue numbers / Feature requests?

Nope.
#### Screenshots (If appropiate)

From the Gitlab service page:

In Asana:

PR #8580, opened 2015-01-15
Add Asana service

j0k3r:
This PR is the following of an other one (which has been closed) #7394 

---

Current state: Ready !
- [x] working asana communication
- [x] ability to restrict to Branch (like the `master`)
#### What does this MR do?

When someone commit with a message including
- "blabla #123456", it will attach a new message to the task 123456 in Asana.
- "fix #456789", it will attach a new message and close (mark as done) the task 456789 in Asana

Also, I've added ability to render `service.help` as markdown. It's more easy to write the help message then.
#### Are there points in the code the reviewer needs to double check?
- `app/models/project_services/asana_service.rb`
- `app/views/projects/services/_form.html.haml`
#### Why was this MR needed?

We use Asana as our Issue tracking system. It will ease the relation between Asana and Gitlab by linking Asana tickets to git commit.
#### What are the relevant issue numbers / Feature requests?

Nope.
#### Screenshots (If appropiate)

From the Gitlab service page:

In Asana:

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -151,6 +151,9 @@ gem "gemnasium-gitlab-service", "~> 0.2"
 # Slack integration
 gem "slack-notifier", "~> 1.0.0"
 
+# Asana integration
+gem 'asana', '~> 0.0.6'
+
 # d3
 gem "d3_rails", "~> 3.1.4"
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -23,6 +23,10 @@ GEM
       activemodel (= 4.1.1)
       activesupport (= 4.1.1)
       arel (~> 5.0.0)
+    activeresource (4.0.0)
+      activemodel (~> 4.0)
+      activesupport (~> 4.0)
+      rails-observers (~> 0.1.1)
     activesupport (4.1.1)
       i18n (~> 0.6, >= 0.6.9)
       json (~> 1.7, >= 1.7.7)
@@ -36,6 +40,8 @@ GEM
       activerecord (>= 2.3.0)
       rake (>= 0.8.7)
     arel (5.0.1.20140414130214)
+    asana (0.0.6)
+      activeresource (>= 3.2.3)
     asciidoctor (0.1.4)
     attr_required (1.0.0)
     awesome_print (1.2.0)
@@ -402,6 +408,8 @@ GEM
       bundler (>= 1.3.0, < 2.0)
       railties (= 4.1.1)
       sprockets-rails (~> 2.0)
+    rails-observers (0.1.2)
+      activemodel (~> 4.0)
     rails_autolink (1.1.6)
       rails (> 3.1)
     rails_best_practices (1.14.4)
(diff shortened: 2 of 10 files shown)


TeatroIO:
I've prepared a stage to preview changes. <a href='http://j0k3r-asana.c68bfdc4cac4c32c41e1a922e73062ea.ttrcloud.com' target='_blank'>Open stage</a> or <a href='http://j0k3r-asana.c68bfdc4cac4c32c41e1a922e73062ea.ttrcloud.com/teatro.log' target='_blank'>view logs</a>.

j0k3r:
I have one problem left.

I can't save the **Restrict to branch** parameter in the service edition. Whatever I put inside this field, nothing is saved.

To test it, you can go [there](http://j0k3r-asana.c68bfdc4cac4c32c41e1a922e73062ea.ttrcloud.com/h5bp/html5-boilerplate/services/asana/edit) (using default login/pass).

Any help will be welcome :grin:

123Haynes:
@j0k3r Just a guess, but it looks like you forgot to add the restrict to branch field in services_controller.rb (around line 44)
The api_key field is already in there so that one works.

j0k3r:
@123Haynes Oh yeah, good catch! Thanks. I will try it

jvanbaarsen:
@j0k3r Can you please make sure the items marked by HoundCI are fixed?

j0k3r:
@jvanbaarsen I've fixed HoundCI reports. The build failed but it isn't related to my changes (IMHO).

j0k3r:
@jvanbaarsen It's ok for me :white_check_mark:

randx:
Looks good to me

casenjo:
Sorry to bring this PR back from the dead, I really like this function but I had a question about where it's placed and how to configure it (or maybe there's something I misunderstood).

Right now it appears at the Project Settings > Services level. The problem I see with this is that each user in Asana has their own Personal Access Token. If I were to configure the integration using my token, then when a teammate makes a commit message, the comment that appears on Asana will be added by my account as opposed to the Asana account that corresponds to them.

This looks a little odd, since a specific user is assigned a task in Asana it should (usually) be that same user that fixes and closes the task.

Is this by design? Would I have to create a special user in Asana to be the "commit bot" or did I miss something?

randx:
> Would I have to create a special user in Asana to be the "commit bot" or did I miss something?

@casenjo I am not familiar with Asana but probably you need kind of bot user which will be used to post information.

j0k3r:
Hello @casenjo,
This service work the same way Github handled it: you need an extra user for Gitlab and then use its personal access token. You need to take care of project access on Asana for this user, otherwise, nothing will show up on Asana.

Each action will be performed by this user on Asana (commenting, closing, etc.).
If we choose to use the personal access token for **each** Gitlab user, it'll be a much harder service to build. Also, it'll be harder to check for each user if (s)he has access to the Asana project, etc.

PR #8602, opened 2015-01-17
Use IID in API

jubianchi:
We this change, users will be able to browse issues, merge requests and milestones APIs using IIDs which feels more natural.

### DIFF
--- a/lib/api/helpers.rb
+++ b/lib/api/helpers.rb
@@ -173,6 +173,10 @@ module API
       end
     end
 
+    def filter_by_iid(items, iid)
+      items.where(iid: iid)
+    end
+
     # error helpers
 
     def forbidden!(reason = nil)
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -20,6 +20,7 @@ v 7.11.0 (unreleased)
   - Don't crash when an MR from a fork has a cross-reference comment from the target project on of its commits.
   - Include commit comments in MR from a forked project.
   - Fix adding new group members from admin area
+  - Query issues, merge requests and milestones with their IID through API (Julien Bianchi)
   - Add default project and snippet visibility settings to the admin web UI.
   -
   - Fix bug where commit data would not appear in some subdirectories (Stan Hu)
--- a/doc/api/issues.md
+++ b/doc/api/issues.md
@@ -99,11 +99,13 @@ GET /projects/:id/issues?labels=foo,bar
 GET /projects/:id/issues?labels=foo,bar&state=opened
 GET /projects/:id/issues?milestone=1.0.0
 GET /projects/:id/issues?milestone=1.0.0&state=opened
+GET /projects/:id/issues?iid=42
 ```
 
 Parameters:
 
 - `id` (required) - The ID of a project
+- `iid` (optional) - Return the issue having the given `iid`
 - `state` (optional) - Return `all` issues or just those that are `opened` or `closed`
 - `labels` (optional) - Comma-separated list of label names
 - `milestone` (optional) - Milestone title
--- a/doc/api/merge_requests.md
+++ b/doc/api/merge_requests.md
@@ -10,11 +10,13 @@ The pagination parameters `page` and `per_page` can be used to restrict the list
 GET /projects/:id/merge_requests
 GET /projects/:id/merge_requests?state=opened
 GET /projects/:id/merge_requests?state=all
+GET /projects/:id/merge_requests?iid=42
 ```
 
 Parameters:
 
 - `id` (required) - The ID of a project
+- `iid` (optional) - Return the request having the given `iid`
 - `state` (optional) - Return `all` requests or just those that are `merged`, `opened` or `closed`
 - `order_by` (optional) - Return requests ordered by `created_at` or `updated_at` fields. Default is `created_at`
 - `sort` (optional) - Return requests sorted in `asc` or `desc` order. Default is `desc`
@@ -388,6 +390,6 @@ Parameters:
 ]
 ```
 
-## Comments on issues
+## Comments on merge requets
 
 Comments are done via the notes resource.
--- a/doc/api/milestones.md
+++ b/doc/api/milestones.md
@@ -6,6 +6,7 @@ Returns a list of project milestones.
 
 ```
 GET /projects/:id/milestones
+GET /projects/:id/milestones?iid=42
 ```
 
 ```json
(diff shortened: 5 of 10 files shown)


TeatroIO:
I've prepared a stage to preview changes. <a href='http://jubianchi-api-iid.c68bfdc4cac4c32c41e1a922e73062ea.ttrcloud.com' target='_blank'>Open stage</a> or <a href='http://jubianchi-api-iid.c68bfdc4cac4c32c41e1a922e73062ea.ttrcloud.com/teatro.log' target='_blank'>view logs</a>.

jvanbaarsen:
Can you please make sure all the HoundCI items are fixed? Also make sure the tests are passing.

jubianchi:
@jvanbaarsen done ;)

jvanbaarsen:
Can you squash the commits? After that i'll mark this as ready for merge :)

jubianchi:
@jvanbaarsen Rebased and squashed ;)

jvanbaarsen:
@jubianchi Since your last rebase there has been some changes on master then prevents your PR from being merged. Can you please make sure its mergeable again?

jubianchi:
@jvanbaarsen done ;)

lib/api/helpers.rb
```diff
@@ -70,6 +70,39 @@ def find_group(id)
       end
     end
 
+    def project_issue
+      begin
+        @issue ||= user_project.issues.find(params[:issue_id])
+      rescue
+        @issue ||= user_project.issues.find_by(iid: params[:issue_id])
+      end
```
NARKOZ: `@issue = user_project.issues.where('id=? OR iid=?', params[:issue_id], params[:issue_id]).first`

lib/api/helpers.rb
```diff
@@ -70,6 +70,39 @@ def find_group(id)
...
+    end
+
+    def project_merge_request
+      begin
+        @merge_request ||= user_project.merge_requests.
+            find(params[:merge_request_id])
+      rescue
+        @merge_request ||= user_project.merge_requests.
+            find_by(iid: params[:merge_request_id])
+      end
```
NARKOZ: ditto

lib/api/helpers.rb
```diff
@@ -70,6 +70,39 @@ def find_group(id)
...
+      @merge_request || not_found!('Merge Request')
+    end
+
+    def project_milestone
+      begin
+        @milestone ||= user_project.milestones.find(params[:milestone_id])
+      rescue
+        @milestone ||= user_project.milestones.
+            find_by(iid: params[:milestone_id])
+      end
```
NARKOZ: same

randx:
I am worried about this PR. Using both id and iid will complicate things in next way: 

* having both `id` and `iid` search can cause unexpected results. Example you ask for issue with iid `3` but get issue with id `3`. This makes writing software for API into nightmare because you can not guarantee correct result. And we can not drop ID support.  
* we need to go through API to be sure we return iid everywhere. Otherwise workflow can be broken.

If second point is solvable but first one is really worry me.

jubianchi:
@randx when building this patch I was also worried by the first point. In fact it can lead to problem:

### Project A

| ID | IID |
|----|----|
| 1 | 1 |
| **4** | 2 |
| 5 | 3 |
| 7 | **4** |

### Project B

| ID | IID |
|----|----|
| **2** | 1 |
| **3** | **2** |
| 6 | **3** |
| 8 | 4 |

This PR will keep the actuel behavior (i.e find by ID first then fallback on IID) but problems will still happen.
The only solution I can think of now to solve this is to add a specific route for IID searches : `/projects/:project_id/issues/iid/:iid` for example.

To me, using IID in the API will really help providing tools to work with gitlab. Actually, we can work around this missing feature by maintaining a map (ID <=> IID) in the tool itself but it requires additional work.

randx:
> add a specific route for IID searches

I am afraid in this case we cause to duplicated - both in code and documentation. Because you basically need to double each route for issue/mr. 

> using IID in the API will really helps providing tools to work with gitlab

I wish it was less painful. `IID` for me is internal id and I never supposed people using it in API routes.  
If you can work around  - it will be great. I appreciate all your work you did here but with Problem N1 this is not  acceptable.

jubianchi:
@randx A friend came up with a solution for the problem we pointed on my patch.

What about making the `iid` a parameter of the `/projects/:project_id/issues` route ? Something like:

```
GET /issues?iid=123
GET /issues?iid=123,456
```

If this seems fine for you, I'll update the PR. If not, I'll try to think of another solution ;)

Thoughts ?

randx:
@jubianchi how much code duplication will we get in this case?

jubianchi:
@randx I think implementing the `GET` parameter filtering will introduce less duplication than the patch actually does.

I think I would add a generic helper for both issues, MRs and milestones, something like:

```ruby
def filter_by_iid(items, iids)
	items.where(iid: iids.split(','))
end
```

And call it in controllers (`issues` for example):

```ruby
issues = filter_by_iid(issues, params[:iid]) unless params[:iid].nil?
```

randx:
@jubianchi but why you need to pass several iids separated by coma?

jubianchi:
@randx In fact, handling multiple IIDs at once is not really required. I did this because the `labels` filter allows the same and I thought it could be useful for bulk requests: the `iid` parameter/filter will be added to the `issues` route which is a listing. It seemed useful to allow users to filter this listing and still get multiple issues at once.

randx:
@jubianchi i am a bot confused. I thought you need iid for getting single issue info. Like /issues/:iid.json

jubianchi:
@randx you are right I only need to fetch a single issue with its IID. 

I'll submit the patch and only support single IID for now ;)

jvanbaarsen:
@jubianchi Does that mean this PR can be closed for now?

jubianchi:
@jvanbaarsen sorry, did not had time to get back to this PR. Can we keep it open for a few days so I can find some time to clear ?

jubianchi:
@jvanbaarsen everything should be ok now ;) sorry for the delay

jvanbaarsen:
@jubianchi Some things have been merged, and now your PR has become unmergeable. Can you please rebase with master?

jubianchi:
Of course I can

jubianchi:
@jvanbaarsen the branch has been rebased ;)

randx:
I understand that is kind of hack passing iid filter to INDEX endpoints to get issues by IID but at least its easy and clean implementation and it does not break compatibility. So I think we can merge it. Thank you @jubianchi

PR #8610, opened 2015-01-19
Fixed assets inclusion due to update of jquery-ui-rails to version 5.0.3

SSPkrolik:
Requirements in `scss` assets were broken with the release of jquery-ui-rails version 5.x, so new syntax for paths was introduced.

Without the edits from the commit, it gave `Sprocket::FileNotFound` exception

### DIFF
--- a/app/assets/javascripts/application.js.coffee
+++ b/app/assets/javascripts/application.js.coffee
@@ -5,7 +5,7 @@
 # the compiled file.
 #
 #= require jquery
-#= require jquery.ui.all
+#= require jquery-ui
 #= require jquery_ujs
 #= require jquery.cookie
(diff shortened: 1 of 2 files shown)


TeatroIO:
I've prepared a stage to preview changes. <a href='http://zeoalliance-master.c68bfdc4cac4c32c41e1a922e73062ea.ttrcloud.com' target='_blank'>Open stage</a> or <a href='http://zeoalliance-master.c68bfdc4cac4c32c41e1a922e73062ea.ttrcloud.com/teatro.log' target='_blank'>view logs</a>.

SSPkrolik:
There was a mistake with running bundle update, that's why the request is not needed.

PR #8007, opened 2014-10-10
Allow HTML tags in user Markdown input

mr-vinn:
This change fixes #7066, and issues [#147](https://gitlab.com/gitlab-org/gitlab-ce/issues/147), [#408](https://gitlab.com/gitlab-org/gitlab-ce/issues/408) and [#476](https://gitlab.com/gitlab-org/gitlab-ce/issues/476) from gitlab.com.

The Markdown parser currently uses Redcarpet's `:filter_html` option, which removes all HTML tags from user input.  The parser also calls ActionView's `#sanitize` method, which removes HTML tags and attributes that are not whitelisted.

GitLab's [Markdown documentation](https://github.com/gitlabhq/gitlabhq/blob/master/doc/markdown/markdown.md#inline-html) says that inline HTML should be allowed, which seems incompatible with the `:filter_html` option.  This change disables that option to allow safe tags from user input to appear in the rendered HTML.

However, relying on `#sanitize` at the end of parsing doesn't handle javascript in tag attributes correctly.  For example, this input:

``` html
<a href="javascript:alert('foo')">link text</a>
```

is rendered as:

``` html
<a href="/namespace101/gitlabhq/blob/markdown/javascript:alert(&#39;foo&#39;)">link text</a>
```

To handle javascript in attributes, I added a call to `#sanitize` before parsing relative links.  This results in the correctly sanitized output:

``` html
<a>link text</a>
```

PR #8620, opened 2015-01-21
Added a note saying inline HTML is disabled by default

namnatulco:
See #8007.

### DIFF
--- a/doc/markdown/markdown.md
+++ b/doc/markdown/markdown.md
@@ -420,6 +420,8 @@ Quote break.
 
 You can also use raw HTML in your Markdown, and it'll mostly work pretty well.
 
+Note that inline HTML is disabled in the default Gitlab configuration, although it is [possible](https://github.com/gitlabhq/gitlabhq/pull/8007/commits) for the system administrator to enable it.
+
 ```no-highlight
 <dl>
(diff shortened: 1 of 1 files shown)


TeatroIO:
I've prepared a stage to preview changes. <a href='http://namnatulco-master.c68bfdc4cac4c32c41e1a922e73062ea.ttrcloud.com' target='_blank'>Open stage</a> or <a href='http://namnatulco-master.c68bfdc4cac4c32c41e1a922e73062ea.ttrcloud.com/teatro.log' target='_blank'>view logs</a>.

jvanbaarsen:
@namnatulco Just to be sure, this PR can only be merged if the PR you refered to has been merged right?

namnatulco:
Quite the opposite! This PR only makes sense if the referred PR is not (yet) merged. The referred PR fixes the problem; this PR writes a note in the documentation saying the problem exists, and how to disable it locally.

I think this PR makes a bit more sense -- in some cases, enabling the interpretation of user-supplied HTML could potentially lead to attacks (which is presumably why it is disabled by default in the first place).

jvanbaarsen:
@namnatulco Thanks :)

Issue #7373, opened 2014-07-25
HipChat requires Oauth2 now


garlandkr:
When adding a token to a room and running the test message there is a resulting error:

```
Started GET "test/services/hipchat/test" for x.x.x.x at 2014-07-25 19:42:04 +0000
Processing by Projects::ServicesController#test as HTML
  Parameters: {"project_id"=>"test", "id"=>"hipchat"}
Completed 500 Internal Server Error in 135ms

HipChat::Unauthorized (Access denied to room `test'):
  app/models/project_services/hipchat_service.rb:45:in `execute'
  app/controllers/projects/services_controller.rb:29:in `test'


Started PUT "test" for x.x.x.x at 2014-07-25 19:42:10 +0000
Processing by Projects::ServicesController#update as HTML
  Parameters: {"utf8"=>"✓", "authenticity_token"=>"test", "service"=>{"active"=>"1", "token"=>"test", "room"=>"test"}, "project_id"=>"test", "id"=>"hipchat"}
Redirected to https://test/services/hipchat/edit
Completed 302 Found in 40ms (ActiveRecord: 6.5ms)
Started GET "test/services/hipchat/edit" for x.x.x.x at 2014-07-25 19:42:10 +0000
Processing by Projects::ServicesController#edit as HTML
  Parameters: {"project_id"=>"test", "id"=>"hipchat"}
Completed 200 OK in 83ms (Views: 45.8ms | ActiveRecord: 5.9ms)
Started GET "test/services/hipchat/test" for x.x.x.x at 2014-07-25 19:42:11 +0000
Processing by Projects::ServicesController#test as HTML
  Parameters: {"project_id"=>"test", "id"=>"hipchat"}
Completed 500 Internal Server Error in 135ms

HipChat::Unauthorized (Access denied to room `test'):
  app/models/project_services/hipchat_service.rb:45:in `execute'
  app/controllers/projects/services_controller.rb:29:in `test'
```

Account credentials need to be added somewhere so an authenticated request can be made from the GitLab settings page for the repo.

jerrac:
Just ran into this issue here at work. Is there a way to tell how many people have subscribed to this issue? If we just look at how many people have commented, there aren't many who are using the HipChat integration...

aangov:
Confirming the same issue. Gitlab is not usable with HipChat anymore.

jamek:
[...]

PR #8645, opened 2015-01-24
Add api version property on hipchat service

chulkilee:
There are two types of Hipchat API token - group-level and account-level. Found that group-level api token is not working with v2 API.

So let's make it configurable.

See also #7373

### DIFF
--- a/app/controllers/projects/services_controller.rb
+++ b/app/controllers/projects/services_controller.rb
@@ -1,5 +1,5 @@
 class Projects::ServicesController < Projects::ApplicationController
-  ALLOWED_PARAMS = [:title, :token, :type, :active, :api_key, :subdomain,
+  ALLOWED_PARAMS = [:title, :token, :type, :active, :api_key, :api_version, :subdomain,
                     :room, :recipients, :project_url, :webhook,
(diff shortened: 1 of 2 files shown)


TeatroIO:
I've prepared a stage to preview changes. <a href='http://chulkilee-hipchat-api-version.c68bfdc4cac4c32c41e1a922e73062ea.ttrcloud.com' target='_blank'>Open stage</a> or <a href='http://chulkilee-hipchat-api-version.c68bfdc4cac4c32c41e1a922e73062ea.ttrcloud.com/teatro.log' target='_blank'>view logs</a>.

chulkilee:
I'm using on my installation and it works well.

stanhu:
I don't think this is necessary. HipChat v1 API is deprecated. GitLab should mandate the use of v2. See https://github.com/gitlabhq/gitlabhq/issues/7373#issuecomment-78201563.

frenck:
While v1 maybe deprecated, I really think this is a choice of the user.

Maybe use a select instead of a text field...

jvanbaarsen:
@chulkilee Can you please make this mergeable again?

chulkilee:
@jvanbaarsen done

PR #8676, opened 2015-01-29
Refactor Repository to use new RepositoryCache class

rspeicher:
Abstracts away the lower-level implementation details from the Repository model.

### DIFF
--- a/lib/repository_cache.rb
+++ b/lib/repository_cache.rb
@@ -1,0 +1,25 @@
+# Interface to the Redis-backed cache store used by the Repository model
+class RepositoryCache
+  attr_reader :namespace
+
+  def initialize(namespace, backend = Rails.cache)
+    @namespace = namespace
+    @backend = backend
+  end
+
+  def cache_key(type)
+    "#{type}:#{namespace}"
+  end
+
+  def expire(key)
+    backend.delete(cache_key(key))
+  end
+
+  def fetch(key, &block)
+    backend.fetch(cache_key(key), &block)
+  end
+
+  private
+
+  attr_reader :backend
+end
--- a/spec/lib/repository_cache_spec.rb
+++ b/spec/lib/repository_cache_spec.rb
@@ -1,0 +1,34 @@
+require 'rspec'
+require_relative '../../lib/repository_cache'
+
+describe RepositoryCache do
+  let(:backend) { double('backend').as_null_object }
+  let(:cache) { RepositoryCache.new('example', backend) }
+
+  describe '#cache_key' do
+    it 'includes the namespace' do
+      expect(cache.cache_key(:foo)).to eq 'foo:example'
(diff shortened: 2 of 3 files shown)


TeatroIO:
I've prepared a stage to preview changes. <a href='http://tsigo-rs-repository-cache.c68bfdc4cac4c32c41e1a922e73062ea.ttrcloud.com' target='_blank'>Open stage</a> or <a href='http://tsigo-rs-repository-cache.c68bfdc4cac4c32c41e1a922e73062ea.ttrcloud.com/teatro.log' target='_blank'>view logs</a>.

spec/lib/repository_cache_spec.rb
```diff
@@ -0,0 +1,34 @@
+require 'rspec'
+require_relative '../../lib/repository_cache'
```
tsigo: Why not just `require 'spec_helper'`? Because it takes *forever* to load if you're just BDD-ing a single class as I was here.

lib/repository_cache.rb
```diff
@@ -0,0 +1,26 @@
+# Interface to the Redis-backed cache store used by the Repository model
+class RepositoryCache
+  attr_reader :namespace
+
+  def initialize(namespace, backend = Rails.cache)
```
tsigo: Why the second argument? [Dependency Injection](http://en.wikipedia.org/wiki/Dependency_injection)

randx:
@tsigo this looks good. Can you make it mergeable and ping me?

tsigo:
@randx Should be good to go.

randx:
@tsigo thank you

lib/repository_cache.rb
```diff
@@ -0,0 +1,25 @@
...
+    backend.delete(cache_key(key))
+  end
+
+  def fetch(key, &block)
+    backend.fetch(cache_key(key), &block)
+  end
+
+  private
+
+  attr_reader :backend
```
randx: @tsigo why this attr is in end of file?

tsigo: I wanted the `attr_reader` for backend but it should be private, which traditionally goes at the bottom. I think you can do `private attr_reader :backend` all on one line and at the top, but that's a bit unusual.

tsigo: Actually, after double-checking, that wasn't doing what I thought it was doing. If you want to make it public and move it back to the top that's fine. Either's good.

randx: I think we should move it on top. I usually dont scroll to end of file to check available attributes

tsigo: Sure thing. Do you want to do that yourself? I think I'd have to submit another PR.

randx: @tsigo I dont have too much time for this now so please submit

PR #8686, opened 2015-01-29
add "Uplaod" and "Replace" functionality

liyakun:
add "replace" functionality, user can replace existing file in current repository.
add "upload" functionality, user can upload new file into current repository.
replace_01
replace_02
upload_01
upload_02


TeatroIO:
I've prepared a stage to preview changes. <a href='http://liyakun-7-5-stable.c68bfdc4cac4c32c41e1a922e73062ea.ttrcloud.com' target='_blank'>Open stage</a> or <a href='http://liyakun-7-5-stable.c68bfdc4cac4c32c41e1a922e73062ea.ttrcloud.com/teatro.log' target='_blank'>view logs</a>.

jvanbaarsen:
@dblessing Would it be possible that you take a look at this and help @liyakun on the way?

dblessing:
@randx before we dive in to review, what do you think of the "replace" and "upload" features?

jvanbaarsen:
@dblessing Good point!

dblessing:
@liyakun At the moment I'm not sure I feel the replace feature is necessary. Do you have a specific use-case that requires this? Advanced git operations are best suited to the CLI, in my opinion. The web editor does allow a user to 'edit' and then copy/paste to 'replace' the file contents.

The upload feature could have some use. However, I think it would be best served as a part of the create new file popup as opposed to it's own button. For example, the popup could have a spot to paste text *or* click to upload a file and commit.

liyakun:
@dblessing The copy and paste works for text file, the replace feature also works for other format file, like image. Also, it is easier to replace than copying and pasting. Maybe it is not necessary, just a consideration for above cases.

I did not find a good position for the upload feature at first,  I think your suggestion is good, you may also ask for other people's opinion.

73:
Hi all,
I think that strategy was to keep alterations to the original behavior to a minimum to keep alienation of users who rely on the copy/paste feature to a minimum.

Looking at:
http://feedback.gitlab.com/forums/176466-general/suggestions/5867835-allow-file-upload-to-repositories-from-the-web-int
Which is another suggestion that support this feature, besides:
http://feedback.gitlab.com/forums/176466-general/suggestions/6961707-create-a-functionality-for-user-to-upload-files-to

I think a "normal" oder "untechy" user like a general manager, customer oder partner (e.g. designers, psychologists, lawyers ...) thinks in files. The user might not want to open a file to submit changes because it was done with a locally installed application (e.g. Gimp, SPSS, R, Word, Excel,...) to generate the changes in the first place.

While a programmer, that just wants to change a line of source code (while having no git client at hand) might want to see the contents of the file.

I think it might be a good idea to keep them separate because of the way different kinds of files are handled by gitlab now. Further I would like to summarise that both ways have their valid use case (or user story if you want). 

I hope this makes any sense to you guys,
73

jvanbaarsen:
@dblessing Did you know if enything was decided about this PR?

73:
Hi,
is there a timeline for this PR yet?
I don't want to sound impatient, but this feature would mean a real game changer in terms of collaboration for us and we would really like to have it in the project.

Thanks for all your hard work,
73

dblessing:
@liyakun Can you please clean up all the hound warnings?

liyakun:
@dblessing, thank you, I already did the code style correction.

randx:
@liyakun can you please rebase on current master so CI includes more tests and also squash your commits into one. After this please make sure CI is green and ping me so I can review.

PR #8687, opened 2015-01-29
Use pry-rails gem instead of pry

rspeicher:
pry-rails has pry as a dependency and this lets us have all that pry goodness when we run `rails console` :heart:

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -221,7 +221,7 @@ group :development, :test do
   gem 'spinach-rails'
   gem "rspec-rails"
   gem "capybara", '~> 2.2.1'
-  gem "pry"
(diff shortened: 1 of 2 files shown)


TeatroIO:
I've prepared a stage to preview changes. <a href='http://tsigo-rs-pry-rails.c68bfdc4cac4c32c41e1a922e73062ea.ttrcloud.com' target='_blank'>Open stage</a> or <a href='http://tsigo-rs-pry-rails.c68bfdc4cac4c32c41e1a922e73062ea.ttrcloud.com/teatro.log' target='_blank'>view logs</a>.

randx:
Awesome!

PR #8730, opened 2015-02-04
Specify shell while run me as git user

drunkard:
Some users disabled "git" user's shell after finished installation, this
will lead to "This account is currently not available" and could not
run /etc/init.d/gitlab, this dirty trick fix it.

Signed-off-by: Drunkard Zhang gongfan193@gmail.com

### DIFF
--- a/lib/support/init.d/gitlab.default.example
+++ b/lib/support/init.d/gitlab.default.example
@@ -29,3 +29,9 @@ web_server_pid_path="$pid_path/unicorn.pid"
 # sidekiq_pid_path defines the path in which to create the pid file for sidekiq
 # The default is "$pid_path/sidekiq.pid"
 sidekiq_pid_path="$pid_path/sidekiq.pid"
+
+# shell_path defines the path of shell for "$app_user" in case you disabled
+# shell of "$app_user" by commands like `usermod -s /sbin/nologin $app_user"
+# for security decision.
+# The default is "/bin/bash"
+shell_path="/bin/bash"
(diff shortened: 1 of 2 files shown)


TeatroIO:
I've prepared a stage to preview changes. <a href='http://drunkard-master.c68bfdc4cac4c32c41e1a922e73062ea.ttrcloud.com' target='_blank'>Open stage</a> or <a href='http://drunkard-master.c68bfdc4cac4c32c41e1a922e73062ea.ttrcloud.com/teatro.log' target='_blank'>view logs</a>.

jvanbaarsen:
I'm not sure if we should support disabling bash for Git user. @randx / @maxlazio What do you think?

jvanbaarsen:
ping @randx

randx:
Looks good to me. @maxlazio any objections?

jvanbaarsen:
ping @maxlazio

maxlazio:
I think this is ok as with manual installation you have an option to change to different shell than bash if you are using a different one. Omnibus uses a different solution than this. @jvanbaarsen This is not recommending that people disable bash, this just allows users to specify which shell they are using.

lib/support/init.d/gitlab.default.example
```diff
@@ -29,3 +29,9 @@ web_server_pid_path="$pid_path/unicorn.pid"
 # sidekiq_pid_path defines the path in which to create the pid file for sidekiq
 # The default is "$pid_path/sidekiq.pid"
 sidekiq_pid_path="$pid_path/sidekiq.pid"
+
+# shell_path defines the path of shell for "$app_user" in case you disabled
+# shell of "$app_user" by commands like `usermod -s /sbin/nologin $app_user"
```
maxlazio: @drunkard I feel that this doesn't belong here the way it was written down now. Can you please change it to:

```
shell_path defines the path of shell for "$app_user" in case you are using 
shell other than "bash"
```

jvanbaarsen:
@maxlazio  I think this one is ready to merge then, right?

jvanbaarsen:
ping @maxlazio

maxlazio:
Thanks

PR #8003, opened 2014-10-09
Fixed gitlab.org issue #633 (Forced all OAuth UserSpec properties to be encoded as UTF-8)

kfazi:
I've found that some omniauth properties are still causing trouble when not forced to be UTF-8.

This commit is forcing every property of Gitlab::OAuth::AuthHash to be UTF-8.

PR #8756, opened 2015-02-08
add common method to force utf8 and force oauth properties to be utf8

zaburt:
OAuth properties cause issues if they include nonascii characters. This change adds a common method to force a string to be UTF-8 and forces all OAuth properties to be UTF-8.

This was discussed in #8003 but the pull request was not completed, here is an updated version.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -27,6 +27,7 @@ v 7.11.0 (unreleased)
   - Move snippets UI to fluid layout
   - Improve UI for sidebar. Increase separation between navigation and content
   - Improve new project command options (Ben Bodenmiller)
+  - Add common method to force UTF-8 and use it to properly handle non-ascii OAuth user properties (Onur Küçük)
   - Prevent sending empty messages to HipChat (Chulki Lee)
   - Improve UI for mobile phones on dashboard and project pages
   - Add room notification and message color option for HipChat
--- a/lib/gitlab/o_auth/auth_hash.rb
+++ b/lib/gitlab/o_auth/auth_hash.rb
@@ -9,11 +9,11 @@ module Gitlab
       end
 
       def uid
-        auth_hash.uid.to_s
+        Gitlab::Utils.force_utf8(auth_hash.uid.to_s)
       end
 
       def provider
-        auth_hash.provider
+        Gitlab::Utils.force_utf8(auth_hash.provider.to_s)
       end
 
       def info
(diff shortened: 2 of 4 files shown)


TeatroIO:
I've prepared a stage to preview changes. <a href='http://zaburt-force-utf8-on-oauth-properties.c68bfdc4cac4c32c41e1a922e73062ea.ttrcloud.com' target='_blank'>Open stage</a> or <a href='http://zaburt-force-utf8-on-oauth-properties.c68bfdc4cac4c32c41e1a922e73062ea.ttrcloud.com/teatro.log' target='_blank'>view logs</a>.

jirutka:
It'd be IMHO better to monkey-patch the force_utf8 method to the String class, instead of using an utility class.

zaburt:
@jirutka thank you for looking into this. If Gitlab devs prefer to do this with monkey patching I can change it, but I find monkey patching String class rather risky in the long term, I prefer not to do it unless it is really necessary. In the current code it seems we don't need to call this method in many places either so I did it this way.

zaburt:
Any updates, comments ?

jvanbaarsen:
I don't like to monkeypatch base classes (like String). @randx what do you feel about that approach?

jirutka:
@jvanbaarsen You can use refinements for safe “monkey-patching” of core classes.

randx:
@jvanbaarsen I am against changing behaviour of core classes. But adding new method like `force_utf8` especially using refinements  can simplify our code a lot

randx:
Again it can be in separate PR.  I think current solution is safe and we can accept it as is.

jvanbaarsen:
@zaburt Can you please fix the issues raised by hound and rebase with master? Ping me when ready so I can merge :)

jirutka:
@randx I can send PR with the refinements approach. You may wanna use [corefines](https://github.com/jirutka/corefines) to preserve fallback for rubies that don’t support refinements (yet).

randx:
@jirutka I though refinements are available since 2.0

randx:
We support only 2.0 and 2.1 - https://github.com/gitlabhq/gitlabhq#requirements

jirutka:
@randx Refinements has been introduced to MRI in 2.0.0, as an experimental feature. However, its design and implementation has been changed then, so refinements in 2.0.x and 2.1+ behaves quite differently.

And what about JRuby? It doesn’t support refinements yet (however, it’ll be soon).

jirutka:
Perhaps differences between 2.0 and 2.1 may not be critical for you, it depends.

jvanbaarsen:
@zaburt We only support MRI, so no need to keep JRuby compatibility in mind :-)

jirutka:
@jvanbaarsen Is there any reason why to disregard JRuby?

jirutka:
Well, if you do care only about MRI 2.0 and 2.1, then I can prepare PR and use refinements in the way compatible with both of these versions.

jirutka:
@randx btw [Corefines](https://github.com/jirutka/corefines) provides also other benefits than just fallback mechanism for backward compatibility. Just for consideration.

zaburt:
@jvanbaarsen I have rebased against master, tests are pssing

randx:
@zaburt thank you

randx:
>  btw Corefines provides also other benefits than just fallback mechanism for backward compatibility. 

We can try that :smile:

Issue #7500, opened 2014-08-14
Photo upload gives 500 with no error log message if cannot save file


JamesHutchison:
This is running GitLab 6.9.2

**Summary:** Gitlab does not provide a helpful error if it cannot save an uploaded photo due to permissions.

**Steps to reproduce:** Set the permissions of the uploads directory (i.e. gitlab/public/uploads) such that  the process cannot write to it.

**Expected Behavior:** Gitlab should serve up a 500 error. The access log should indicate the 500 error, and the error log should provide a helpful error message.

**Observed Behavior:** Gitlab returns 500, the access log indicates the 500 error. The error log provides no message.

Issue #7052, opened 2014-05-30
404 Image Upload on Issue


MrKeiKun:
**Summary:**
Whenever I upload an image and posted it an issue(possible even in comment), it doesn't show the image that i uploaded.

**Logs:**
```
Started GET "/uploads/nette/katana/25857b313e/rawr18.png" for 112.207.110.70 at 2014-05-31 00:00:17 +0800
Processing by GroupsController#show as HTML
Parameters: {"id"=>"uploads/nette/katana/25857b313e/rawr18.png"}
Filter chain halted as :authorize_read_group! rendered or redirected
Completed 404 Not Found in 6ms (Views: 0.7ms | ActiveRecord: 0.7ms)
```

**ScreenShot:**
![rawr20](https://cloud.githubusercontent.com/assets/4362134/3132855/afab68f6-e813-11e3-8925-6134f19f6049.png)

**Related to:**
https://groups.google.com/forum/#!topic/gitlabhq/dNgZ9_wL3GA

Razer6:
Thanks for the issue report. This issue has already been fixed in newer versions of GitLab. Due to the size of this project and our limited resources we are only able to support the latest stable release as outlined in our [contributing guidelines](https://gitlab.com/gitlab-org/gitlab-ce/blob/master/CONTRIBUTING.md#issue-tracker). In order to get this bug fix and enjoy many new features please [upgrade](https://gitlab.com/gitlab-org/gitlab-ce/tree/master/doc/update). If you still experience issues at that time please open a new issue following our issue tracker guidelines found in the [contributing guidelines](https://gitlab.com/gitlab-org/gitlab-ce/blob/master/CONTRIBUTING.md#issue-tracker-guidelines).

Issue #6281, opened 2014-02-09
Group Avatar not working


MrKeiKun:
**Summary:**
For some reasons, I can't see the group avatar i uploaded.
I check the image via right click and open image to another tab
the link given was `http://example.com/uploads/group/avatar/5/megane.jpg`.
then I checked my gitlab server and went to the path the link says. And this what I found
```
[git@Neko 5]$ pwd
/home/git/gitlab/public/uploads/group/avatar/5
[git@Neko 5]$ ls
megane.jpg
[git@Neko 5]$
```
**Screenshots:**![rawr45](https://f.cloud.github.com/assets/4362134/2119766/2d50e6e8-9187-11e3-83ca-207c228aab62.png)
![rawr46](https://f.cloud.github.com/assets/4362134/2119790/e9816bec-9189-11e3-8473-72e4c154830c.png)

**Logs:**
```
112.207.7.177 - - [09/Feb/2014:20:41:18 +0800] "GET /uploads/group/avatar/5/megane.jpg HTTP/1.1" 404 357
```

jpscharf:
I was able to resolve the group avatar issue by updating my nginx config so the root folder was set correctly. Nginx was not basically failing at all requests, including request for static files and forwarding them to the ruby.

Razer6:
Closing for now because seems to be fixed

PR #8769, opened 2015-02-10
check upload dir permissions

bbodenmiller:
Detect permission and ownership issues with the users uploads dir. This helps troubleshoot and correct [omnibus issue 311](https://gitlab.com/gitlab-org/omnibus-gitlab/issues/311). Fixes https://github.com/gitlabhq/gitlabhq/issues/7500, https://github.com/gitlabhq/gitlabhq/issues/7052. Related to https://github.com/gitlabhq/gitlabhq/issues/6281.

### DIFF
--- a/lib/tasks/gitlab/check.rake
+++ b/lib/tasks/gitlab/check.rake
@@ -22,6 +22,7 @@ namespace :gitlab do
       check_gitlab_config_not_outdated
       check_log_writable
       check_tmp_writable
+      check_upload_perms
       check_init_script_exists
       check_init_script_up_to_date
       check_projects_have_namespace
@@ -270,6 +271,29 @@ namespace :gitlab do
         try_fixing_it(
           "sudo chown -R gitlab #{tmp_path}",
           "sudo chmod -R u+rwX #{tmp_path}"
+        )
+        for_more_information(
+          see_installation_guide_section "GitLab"
+        )
+        fix_and_rerun
+      end
+    end
+
+    def check_upload_perms
+      print "Upload directory permissions correct? ... "
+
+      upload_path = Rails.root.join(Gitlab.config.gitlab.user_home,"gitlab-rails","uploads")
+      upload_path_tmp = Rails.root.join(upload_path,"tmp")
+
+      # if tmp upload dir has incorrect permissions, assume others do as well
+      if File.stat(upload_path_tmp).mode == 16877 && File.owned?(upload_path_tmp) # verify drwxr-xr-x permissions
+        puts "yes".green
+      else
+        puts "no".red
+        try_fixing_it(
+          "sudo chown -R #{gitlab_user} #{upload_path}",
+          "sudo find #{upload_path} -type f -exec chmod g+r,o+r {} \\;",
+          "sudo find #{upload_path} -type d -not -path #{upload_path} -exec chmod g+rx,o+rx {} \\;"
         )
         for_more_information(
           see_installation_guide_section "GitLab"


TeatroIO:
I've prepared a stage to preview changes. <a href='http://bbodenmiller-check-upload-perms.c68bfdc4cac4c32c41e1a922e73062ea.ttrcloud.com' target='_blank'>Open stage</a> or <a href='http://bbodenmiller-check-upload-perms.c68bfdc4cac4c32c41e1a922e73062ea.ttrcloud.com/teatro.log' target='_blank'>view logs</a>.

bbodenmiller:
Failure looks to be unrelated.

lib/tasks/gitlab/check.rake
```diff
@@ -278,6 +279,29 @@ namespace :gitlab do
       end
     end
 
+    def check_upload_perms
+      print "Upload directory permissions correct? ... "
+
+      upload_path = Rails.root.join(Gitlab.config.gitlab.user_home,"gitlab-rails","uploads")
+      upload_path_tmp = Rails.root.join(upload_path,"tmp")
```
jacobvosmaer: I think a better way to find the uploads path might be to take `File.realpath(Rails.root.join('public/uploads'))`.

lib/tasks/gitlab/check.rake
```diff
@@ -278,6 +279,29 @@ namespace :gitlab do
...
+
+      # if tmp upload dir has incorrect permissions, assume others do as well
+      if File.stat(upload_path_tmp).mode == 16877 && File.owned?(upload_path_tmp) # verify drwxr-xr-x permissions
+        puts "yes".green
+      else
+        puts "no".red
+        try_fixing_it(
+          "sudo chown -R #{gitlab_user} #{upload_path}",
```
jacobvosmaer: Shouldn't there be a `chmod` on the upload path too here?

lib/tasks/gitlab/check.rake
bbodenmiller: I'll give it a try. As I'm new to Ruby could you explain why it might be better?

lib/tasks/gitlab/check.rake
bbodenmiller: I believe so. Is there also an easy way I can check who should own the upload folder? Omnibus makes this easy I believe just not sure about non-omnibus.

lib/tasks/gitlab/check.rake
jacobvosmaer: Thanks for asking @bbodenmiller . I think I was a little rushed above :)

I think the intended use of 'Rails.root.join' is to find the path to files/directories under either `/home/git/gitlab` or `/opt/gitlab/service/embedded/gitlab-rails`. GitLab the Rails app expects the uploads to be in `public/uploads` under the Rails root, hence `Rails.root.join('public/uploads')`. Now we know that with the omnibus packages, `public/uploads` is actually a symlink. By using `File.realpath` we can find out what it points to.

The gain of all this is that we are no longer making assumptions about the file layout beyond 'uploads are in `public/uploads`'.

jacobvosmaer: For the second variable, I would write `File.join(upload_path, 'tmp')`. That path manipulation does not care about the Rails root.

lib/tasks/gitlab/check.rake
jacobvosmaer: It should be owned by the same user that runs the Unicorn and Sidekiq processes. I think that is `gitlab_user` in this case, right?

lib/tasks/gitlab/check.rake
```diff
@@ -278,6 +279,29 @@ namespace :gitlab do
...
     end
 
+    def check_upload_perms
+      print "Upload directory permissions correct? ... "
+
+      upload_path = Rails.root.join(Gitlab.config.gitlab.user_home,"gitlab-rails","uploads")
+      upload_path_tmp = Rails.root.join(upload_path,"tmp")
+
+      # if tmp upload dir has incorrect permissions, assume others do as well
+      if File.stat(upload_path_tmp).mode == 16877 && File.owned?(upload_path_tmp) # verify drwxr-xr-x permissions
```
jacobvosmaer: Maybe we should skip this test if upload_path_tmp does not exist. (Does it exist on a fresh install?)

For readability, it might also be nice to write `040755` instead of `16877`.

lib/tasks/gitlab/check.rake
jacobvosmaer: I just realized the chmod is done by `find` below. Sorry.

lib/tasks/gitlab/check.rake
bbodenmiller: Awh. I did not realize omnibus is a symlink. I assumed it was not hence why I did it this way. I will check the other way this weekend.

lib/tasks/gitlab/check.rake
bbodenmiller: Good point.

lib/tasks/gitlab/check.rake
bbodenmiller: Actually still a valid point as the `find chmod` only covers subdirectories under `upload_path` and the higher level has both different permissions and group ownership. In my case I didn't have this issue so didn't build a test for it.

bbodenmiller: The `upload_path` is owned by group specified at https://github.com/gitlabhq/omnibus-gitlab/blob/master/files/gitlab-config-template/gitlab.rb.template#L307 from what I can tell. I'm just not sure how to access that variable/where it is in non-omnibus.

jacobvosmaer: The ownership of upload_path is not the same between source installations and omnibus installations. The ownership of uploads_path is the one thing `gitlab-ctl reconfigure` _will_ fix for you. I think it would be sane to not check it, as the code does in this MR.

bbodenmiller: What about non-omnibus though? Are you suggesting changing the ownership checking in this MR or just saying the way it is now is sufficient?

jacobvosmaer: I think the way it is now is sufficient.

jvanbaarsen:

PR #8772, opened 2015-02-10
Fix small grammar issue on projects show page

awiddersheim:
When viewing a project owned by a group it currently looks like this:

image

This commit will change it so it is now the following:

image

Projects owned by individual users will stay the same:

image

### DIFF
--- a/app/views/projects/show.html.haml
+++ b/app/views/projects/show.html.haml
@@ -59,7 +59,7 @@
             %span.light Created on
             #{@project.created_at.stamp('Aug 22, 2013')}
           %p
-            %span.light Owned by
+            %span.light Owned by #{@project.group ? "the" : nil}
             - if @project.group
               #{link_to @project.group.name, @project.group} group
             - else


TeatroIO:
I've prepared a stage to preview changes. <a href='http://awiddersheim-fix-grammar-projects-show.c68bfdc4cac4c32c41e1a922e73062ea.ttrcloud.com' target='_blank'>Open stage</a> or <a href='http://awiddersheim-fix-grammar-projects-show.c68bfdc4cac4c32c41e1a922e73062ea.ttrcloud.com/teatro.log' target='_blank'>view logs</a>.

dserodio:
Nice :+1:

app/views/projects/show.html.haml
```diff
@@ -59,7 +59,7 @@
             %span.light Created on
             #{@project.created_at.stamp('Aug 22, 2013')}
           %p
-            %span.light Owned by
+            %span.light Owned by #{@project.group ? "the" : nil}
             - if @project.group
```
randx: Maybe add the here instead?

awiddersheim: I wanted to continue having `%span.light` applied to the text. Doing it below removed it and didn't make things look very good.

randx: Ok. Thanks

PR #8831, opened 2015-02-20
Fix user API tests succeeding for the wrong reason

bladealslayer:
The requests were missing multiple required attributes, while testing for single attribute missing.
Added test for checking username as required attribute.

### DIFF
--- a/spec/requests/api/users_spec.rb
+++ b/spec/requests/api/users_spec.rb
@@ -110,17 +110,22 @@ describe API::API, api: true  do
     end
 
     it 'should return 400 error if name not given' do
-      post api('/users', admin), email: 'test@example.com', password: 'pass1234'
+      post api('/users', admin), attributes_for(:user).except(:name)
       expect(response.status).to eq(400)
     end
 
     it 'should return 400 error if password not given' do
-      post api('/users', admin), email: 'test@example.com', name: 'test'
+      post api('/users', admin), attributes_for(:user).except(:password)
       expect(response.status).to eq(400)
     end
 
-    it "should return 400 error if email not given" do
-      post api('/users', admin), password: 'pass1234', name: 'test'
+    it 'should return 400 error if email not given' do
+      post api('/users', admin), attributes_for(:user).except(:email)
+      expect(response.status).to eq(400)
+    end
+
+    it 'should return 400 error if username not given' do
+      post api('/users', admin), attributes_for(:user).except(:username)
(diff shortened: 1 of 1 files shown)


TeatroIO:
I've prepared a stage to preview changes. <a href='http://bladealslayer-fix-api-user-tests.c68bfdc4cac4c32c41e1a922e73062ea.ttrcloud.com' target='_blank'>Open stage</a> or <a href='http://bladealslayer-fix-api-user-tests.c68bfdc4cac4c32c41e1a922e73062ea.ttrcloud.com/teatro.log' target='_blank'>view logs</a>.

Razer6:
@jvanbaarsen Can you take a look?

jvanbaarsen:
@bladealslayer What do you think about just putting all in one test case like this:

```
it 'should return 400 error if not all required attributes are given' do
end
```

We already test the validations in `models/user.rb`. That way we are testing behaviour (i.e. does it check for valid attributes and then send back a 400) instead of actual implementation.

bladealslayer:
@jvanbaarsen Not sure I follow. If it would be only one test case, which attribute should be omitted from the request to cause it to fail? Or you mean something like:

```
    it 'should return 400 error if requred attribute not given' do
      for attribute in [:name, :password, :email, :username] do
        post api('/users', admin), attributes_for(:user).delete(attribute)
        expect(response.status).to eq(400)
      end
    end
```

Razer6:
@jvanbaarsen Can you answer?

spec/requests/api/users_spec.rb
```diff
@@ -110,17 +110,22 @@
     end
 
     it 'should return 400 error if name not given' do
-      post api('/users', admin), email: 'test@example.com', password: 'pass1234'
+      post api('/users', admin), attributes_for(:user).delete(:name)
```
NARKOZ: Tests are still wrong, because `delete` returns the value of the deleted key, not a modified `Hash`. 
You should use the `except` method.

jvanbaarsen: @bladealslayer Can you please fix the comment raised by @NARKOZ ?

bladealslayer: Thanks, @NARKOZ Will fix.

bladealslayer: @jvanbaarsen @NARKOZ fixed!

NARKOZ:
Thank you!

PR #8851, opened 2015-02-24
[bugfix] gitlab issue tracker link

cao:
Fixes the incorrect link to the Gitlab issue tracker for non-root URL installations.

### DIFF
--- a/app/models/project_services/gitlab_issue_tracker_service.rb
+++ b/app/models/project_services/gitlab_issue_tracker_service.rb
@@ -27,11 +27,11 @@ class GitlabIssueTrackerService < IssueTrackerService
   end
 
   def project_url
-    namespace_project_issues_path(project.namespace, project)
+    "#{Gitlab.config.gitlab.url}#{namespace_project_issues_path(project.namespace, project)}"
   end
 
   def new_issue_url
-    new_namespace_project_issue_path namespace_id: project.namespace, project_id: project
+    "#{Gitlab.config.gitlab.url}#{new_namespace_project_issue_path namespace_id: project.namespace, project_id: project}"
   end
 
   def issue_url(iid)


TeatroIO:
I've prepared a stage to preview changes. <a href='http://cao-master.c68bfdc4cac4c32c41e1a922e73062ea.ttrcloud.com' target='_blank'>Open stage</a> or <a href='http://cao-master.c68bfdc4cac4c32c41e1a922e73062ea.ttrcloud.com/teatro.log' target='_blank'>view logs</a>.

Razer6:
Already fixed on master

cao:
It would be good to have the repos here and gitlab.com actually sync regularly, or at least designate a repo to work with in "Contributing to Gitlab." Instead, it explicitly states that both are fine to use, yet they are (regularly) out of sync. This is really horrible for contributors because they don't have a clear state to start working with/on.
I cloned the repo about an hour ago and it was not fixed in master (on GitHub) at the time. The last commit to the file was 10/12 days ago. Now, it was fixed *5* hours ago because someone finally decided to push to GitHub and, in turn, sync the repo. Selectively pushing to one or the other repo (and not syncing them automatically) is a bad idea if you accept contributions at both, it should really be changed.

Razer6:
@cao There is another a company internal development server. This particular fix was announced in the [regression issue](https://gitlab.com/gitlab-org/gitlab-ce/issues/1100#note_872321) and has been merged 2 hours ago and then immediately was pushed to github.com as well as gitlab.com.

cao:
It is simply not true that it was synced immediately.

I forked the repo at (taken from GitHub's security log):
    created_at	2015-02-23 22:39:13 -0800

Yet, the commit that fixes the regression is from:
     Tue Feb 24 05:50:32 2015 +0000
which is:
     Mon Feb 23 21:50:32 2015 -0800

The commit predates my fork by nearly 50 minutes, thus, the commit must be in my fork if it was immediately pushed to github.com. It is not in [my fork](/cao/gitlabhq) though, hence, the repos were out of sync for at least 50 minutes.

Furthermore, you are making the my case for me: I need to check two issue trackers and repositories to figure out if a bug has been fixed or if it is being discussed already, this is extremely annoying.

Razer6:
You are rigjht. There was a difference of an hour.

When having a release we always create a meta-issue related to bugs to this version. This is always on gitlab.com since this is the canonical source. This meta-issue is also announced on Twitter with the release candidate. This is created with the first release candidate. While releasing and after releasing we're linking related issues/fixes for this version to this meta-issue. So when seing a new issue after upgrading, looking into this meta-issue might be the best place to investigate if someone is already working on a fix for it. Or if not, link the new problem to this issue so someone can check if this is a regression and needs to be fixed as part of a patch releasr.

PR #8858, opened 2015-02-24
Update version sorter to 2.0.0, fixes #8572

### LINKED ISSUE #8572: The old version of VersionSorter may cause the encoding error in file page.
There is no restriction for the lowest version of VersionSorter in Gemfile, so in current version (1.1.0) it will cause an encoding error when sorting Tags, which turns UTF-8  to ASCII, then trigger a encoding error in file page.

The problem encounter at line #95 in app/helpers/application_helper.rb
https://github.com/Shanicky/gitlabhq/blob/master/app/helpers/application_helper.rb#L95
When we removed the sorting for tags, there is no encoding error in file page

Razer6:
This updates the `version_sorter` gem to 2.0.0. The API is the same, but the implementation improved.
This also includes a fix for encoding. The the current used version, tags get reencoded to ASCII rather than preserving its encoding.

/cc @tsigo

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -608,7 +608,7 @@ GEM
       raindrops (~> 0.7)
     unicorn-worker-killer (0.4.2)
       unicorn (~> 4)
-    version_sorter (1.1.0)
+    version_sorter (2.0.0)
     virtus (1.0.1)
       axiom-types (~> 0.0.5)
       coercible (~> 1.0)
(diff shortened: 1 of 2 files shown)


TeatroIO:
I've prepared a stage to preview changes. <a href='http://razer6-update-version-sorter.c68bfdc4cac4c32c41e1a922e73062ea.ttrcloud.com' target='_blank'>Open stage</a> or <a href='http://razer6-update-version-sorter.c68bfdc4cac4c32c41e1a922e73062ea.ttrcloud.com/teatro.log' target='_blank'>view logs</a>.

tsigo:
:+1:

Razer6:
@tsigo Done so. Added a UTF-8 character.

jvanbaarsen:
@Razer6 I've restarted the build in order to see if it becomes green :)

Razer6:
@jvanbaarsen :green_heart:

jvanbaarsen:
@Razer6 Thanks!

PR #8928, opened 2015-03-06
use constant-time string compare for internal api authentication

Mic92:
Ruby str_equal uses memcmp internally to compare String.
Memcmp is vunerable to timing attacs because it returns early
on mismatch (on most x32 platforms memcmp uses a bytewise comparision).
Devise.secure_compare implements a constant time comparision instead.

### DIFF
--- a/lib/api/helpers.rb
+++ b/lib/api/helpers.rb
@@ -83,7 +83,10 @@ module API
     end
 
     def authenticate_by_gitlab_shell_token!
-      unauthorized! unless secret_token == params['secret_token'].try(:chomp)
+      input = params['secret_token'].try(:chomp)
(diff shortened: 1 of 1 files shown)


TeatroIO:
I've prepared a stage to preview changes. <a href='http://mic92-master.c68bfdc4cac4c32c41e1a922e73062ea.ttrcloud.com' target='_blank'>Open stage</a> or <a href='http://mic92-master.c68bfdc4cac4c32c41e1a922e73062ea.ttrcloud.com/teatro.log' target='_blank'>view logs</a>.

Mic92:
This attack is unpractical over the internet, but might be possible in shared hosting environments or on the same lan.

Razer6:
:+1:

PR #8974, opened 2015-03-19
Add error message when have error on profile screen

duduribeiro:
Like described in issue #8972, when trying to update the profile with an email already in use, the error message only provides the error: `Failed to update profile`. With this change, the message will describe the exactly error occurred in the process.

Before:
update_profile_befor
After:
update_profile_after

PR #8981, opened 2015-03-20
Fix broken paths

duduribeiro:
When I rebased the PR #8974, some tests broken with a invalid path.

This PR fix the paths to use the namespace.


duduribeiro:
I think this is not valid since @randx fixed it in commit 479020e

randx:
@duduribeiro thank you for participating :smile:

Issue #8286, opened 2014-11-11
Green "Accept merge request" is dangerous


coder-hugo:
Since a few months we are using GitLab at work to mange our projects. For reviewing the code we use the merge request feature which is very helpful. But it happened more than once that the reviewer hit accidentally the button "Accept merge request" after modifying the assignee or the milestone. This happens because of several UX issues.
1. You won't get notified that your changes regarding the assignee or the milestone have been saved automatically.
2. There is no confirmation dialog before performing the auto-merge.
3. The color of button is green. As the action behind this button usually will modify your master the button should be red or similar to give the user a fair waring that the action of this button is "dangerous".

IMHO the easiest solution for this issues would be the color change of the button. But I think that a confirmation dialog for such an action should be essential.

Brunozrk:
@rumpelsepp +1

A simple confirmation should save much time!

randx:
Coloring button based on CI status makes sense for me

jvanbaarsen:
We are currently in the process of closing down the issue tracker on GitHub, to prevent duplication with the GitLab.com issue tracker. Since this is an older issue I'll be closing this for now. If you think this is still an issue I encourage you to open it on the [GitLab.com issue tracker](https://gitlab.com/gitlab-org/gitlab-ce/issues).

PR #8983, opened 2015-03-20
Add confirmation when accept merge request

duduribeiro:
Like proposed on issue #8286, this PR implements a simple confirmation dialog to avoid accidentally clicks on `Accept Merge Request` button.

print

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -57,6 +57,7 @@ v 7.10.0 (unreleased)
   - Archive repositories in background worker.
   - Import GitHub, Bitbucket or GitLab.com projects owned by authenticated user into current namespace.
   - Fix and improve help rendering (Sullivan Sénéchal)
+  - Add confirmation dialog when click in "Accept Merge Request".
 
 
 v 7.9.2
@@ -70,9 +71,6 @@ v 7.9.1
   - Fix missing events and in admin Slack service template settings form (Stan Hu)
   - Don't show commit comment button when user is not signed in.
   - Downgrade gemnasium-gitlab-service gem
-
-v 7.9.0
-  - Add confirmation dialog when click in "Accept Merge Request".
 
 v 7.9.0 (unreleased)
   - Add HipChat integration documentation (Stan Hu)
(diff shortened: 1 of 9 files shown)


rumpelsepp:
Personally I would provide a better message. Like "This will merge fork/branch-xyz into master. Are you sure?"

Brunozrk:
:clap:

fpgentil:
:+1: Great, I have many times merged MRs unintentionally :see_no_evil:

duduribeiro:
@rumpelsepp like you proposed:
image

it is ok?

thaaaanks =D

:smile:

Razer6:
:+1: for the idea, but :-1: for the dialog. I hate browser dialogs ;-) Could this be integrated in the GitLab UI?

duduribeiro:
@Razer6, what about this confirmation message? what you think about the confirmation message?
merge_message

thaaanks

randx:
I merge ~ 20 requests per day. It absolutely unacceptable see this red message and type something in input box. There is NOTHING bad in merging branch - its revert-able operation. I dont think it should be any kind of confirmation for merge.

jvanbaarsen:
@duduribeiro I disagree with this approach. This would make merging an awful thing to do (as @randx points out). I'll close this PR for now. If you think it should be implemented I urge you to create a ticket on feedback.gitlab.com so it can be discussed.

Thanks for your efforts though! :heart:

brunozrk:
I think a simple confirmation should be enough (without the input, only a text confirmation). I agree that is revert-able operation, but a confirmation can avoid unintentionally merges and isn't an awful thing to do (just confirm or not :smile:).

Maybe define it as a configuration option in settings?

randx:
> Maybe define it as a configuration option in settings?

You dont want application like this, right? :smile:
 
controls

fvztdk:
That was a great ideia :cry:  someone already opened a ticket on the feedback system ?

Razer6:
A confirmation like GitHub does right now would be ok for me. But a popup would be imho annoying

randx:
I agree @Razer6

PR #8985, opened 2015-03-20
Extend push_tag event to include tag message and last commit

ayufan:


vsizov:
@ayufan Can you fix tests?

related https://gitlab.com/gitlab-org/gitlab-ci/issues/90

ayufan:
Updated. Should be ok.

app/services/git_tag_push_service.rb
```diff
@@ -2,22 +2,38 @@ class GitTagPushService
   attr_accessor :project, :user, :push_data
 
   def execute(project, user, oldrev, newrev, ref)
-    @project, @user = project, user
-    
-    @push_data = build_push_data(oldrev, newrev, ref)
+    if Gitlab::Git.tag_ref?(ref)
```
randx: This is tag push service. This code should be executed only on tag push. What sense to check if it is a tag here?

ayufan: I agree. This is not required, but I'm following here GitPushService where such check is done.

randx: @ayufan ah ok. In this case we need to remove check from GitPushService too

ayufan: So should I remove it from here?

randx: yes. ANd from GitPushService too

ayufan:
Updated

randx:
@ayufan looks good!

ayufan:
Thanks

PR #8988, opened 2015-03-20
Fix GitLab shell setup spacing

atomaka:
Running the `gitlab:shell:setup` task can result in a misalignment.  As seen in the following screenshot:

gitlab-shell-setup-task

After printing the "status dots," a new line should be added to ensure the command prompt is printed at the expected location.  This is a minor issue with a simple fix.


Razer6:
Thanks

PR #8996, opened 2015-03-22
Fix Node.js Error in Documentation

vinkla:
Installing `nodejs` didn't work for me, instead I've installed it but just writing `node`.

### DIFF
--- a/doc/update/7.8-to-7.9.md
+++ b/doc/update/7.8-to-7.9.md
@@ -42,6 +42,8 @@ sudo -u git -H git checkout v2.6.0
 
 ### 4. Install libs, migrations, etc.
 
+Please refer to the [Node.js setup documentation](https://github.com/joyent/node/wiki/installing-node.js-via-package-manager#debian-and-ubuntu-based-linux-distributions) if you aren't running default GitLab server setup.
+
 ```bash
 sudo apt-get install nodejs
 


vinkla:
Maybe better would be to add the example as below. [As in documentation](https://github.com/joyent/node/wiki/installing-node.js-via-package-manager#debian-and-ubuntu-based-linux-distributions).
````
apt-get install curl
curl -sL https://deb.nodesource.com/setup | bash -

**Comment by chulkilee** [comment]
What didn't work for you? I don't think gitlab requires the latest node on 0.10 (from nodesource). If the official package on main repo works, then this should be avoided to add 3rd party repo.

**Comment by vinkla** [comment]
After running `sudo apt-get install nodejs` it returned an error saying it couldn't find the nodejs package. It worked after I followed the installation guide from the Node.js documentation.

**Comment by chulkilee** [comment]
It's on universe - probably you didn't add it to sourcelist. At least the doc should be updated about it.

See [nodejs package page for 14.04](http://packages.ubuntu.com/trusty/nodejs). Also note that this is 0.10.25, which is far behind the latest version.

**Comment by vinkla** [comment]
What about if you are running Debian? Which is the preferred OS?

**Comment by chulkilee** [comment]
For debian, see [this search result](https://packages.debian.org/search?keywords=nodejs) - looks like nodejs is available to debian too.

**Comment by vinkla** [comment]
Weird that it didn't work then. Do I've an old setup maybe?

**Comment by chulkilee** [comment]
Are you using wheezy? nodejs is available in wheezy-backports and I'm not sure it's on it by default.

**Comment by vinkla** [comment]
This is my `/etc/apt/sources.list` file.
```
deb http://ftp.se.debian.org/debian/ wheezy main
deb-src http://ftp.se.debian.org/debian/ wheezy main

deb http://security.debian.org/ wheezy/updates main
deb-src http://security.debian.org/ wheezy/updates main

# wheezy-updates, previously known as 'volatile'
deb http://ftp.se.debian.org/debian/ wheezy-updates main
deb-src http://ftp.se.debian.org/debian/ wheezy-updates main
```

**Comment by chulkilee** [comment]
add following, run 'apt-get update' and retry installing nodejs

```
deb http://ftp.se.debian.org/debian/ wheezy-backports main
deb-src http://ftp.se.debian.org/debian/ wheezy-backports main
```

vinkla:
Does this come out of the box in newer distributions of Debian? Should it be there by default?

randx:
@maxlazio do you think we should merge this?

maxlazio:
@randx I don't think so. I know it would be useful for users to have it one place but then we have to write down what is the process for CentOS 6 and 7 and it can be found out with one search.

vinkla:
Maybe this should be closed then. If anyone else encounter this issue they can use this PR as reference. Perhaps we could just add a link in the documentation? Something like "please refer to the Node.js setup documentation if you aren't running default GitLab server setup". What do you think?

maxlazio:
@vinkla That is a good idea

vinkla:
@maxlazio Thanks. 

Where do we go from here? Do you guys want me to update this commit?

maxlazio:
@vinkla If you do that ping me so I can merge it :)

vinkla:
@maxlazio done :)

maxlazio:
Awesome, thanks!

vinkla:
Thanks :)

PR #9007, opened 2015-03-23
Complete transition to using color_field for selecting colors

atomaka:
The labels color field was updated to use the `color_field` form helper in commit 7606b93.  This PR updates the broadcast messages form to do the same.

### DIFF
--- a/app/views/admin/broadcast_messages/index.html.haml
+++ b/app/views/admin/broadcast_messages/index.html.haml
@@ -21,13 +21,11 @@
   .form-group.js-toggle-colors-container.hide
     = f.label :color, "Background Color", class: 'control-label'
     .col-sm-10
-      = f.color_field :color, placeholder: "#AA33EE", class: "form-control"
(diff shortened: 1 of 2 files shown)


Razer6:
@atomaka Thanks for this PR. Is the hint ` 6 character hex values starting with a # sign.` still needed now?

atomaka:
@Razer6 The other commit did not remove the hint so I left it for consistency.  I'm not sure if there is still a purpose.

Razer6:
You can remove them. In both cases for consistency.

atomaka:
I've removed the help texts from the broadcast messages forms and updated it in the labels creation form.  I also changed from setting the placeholder attribute on the form element to setting the value.  Placeholder is not relevant with color type fields.

Are there other instances of color selection that might benefit from being switched to use the color type input field?

randx:
@atomaka i think its only broadcast messages and labels. But labels already switched. THank you

PR #9008, opened 2015-03-23
Fix generating SSH key fingerprints with OpenSSH 6.8.

sstanovnik:
OpenSSH 6.8 [introduces](http://www.openssh.com/txt/release-6.8) a new feature that changes the default fingerprint format and algorithm used by `ssh-keygen`. This breaks adding new SSH keys, because GitLab expects the colon-delimited format. 

The message the user sees on the Add an SSH Key screen is "Fingerprint cannot be generated", similar to #7413, but the underlying cause is different. 

This change checks the OpenSSH version and explicitly specifies the previous format if needed.

### DIFF
--- a/app/models/key.rb
+++ b/app/models/key.rb
@@ -83,14 +83,14 @@ class Key < ActiveRecord::Base
 
     cmd_status = 0
     cmd_output = ''
+    explicit_fingerprint_algorithm = false
     Tempfile.open('gitlab_key_file') do |file|
       file.puts key
       file.rewind
-      
+
       # OpenSSH 6.8 introduces a new default output format for fingerprints.
       # Check the version and decide which command to use.
       version_output, version_status = popen(%W(ssh -V))
-      explicit_fingerprint_algorithm = false
       if version_status.zero?
         out, _ = version_output.scan /.*?(\d)\.(\d).*?,/
         major, minor = out[0], out[1]
@@ -107,8 +107,15 @@ class Key < ActiveRecord::Base
     end
 
     if cmd_status.zero?
-      cmd_output.gsub /(\h{2}:)+\h{2}/ do |match|
-        self.fingerprint = match
+      if explicit_fingerprint_algorithm
+        cmd_output.gsub /(MD5:)(\h{2}:)+\h{2}/ do |match|
+          match.slice! /^MD5:/
+          self.fingerprint = match
+        end
+      else
+        cmd_output.gsub /(\h{2}:)+\h{2}/ do |match|
+          self.fingerprint = match
+        end
       end
     end
   end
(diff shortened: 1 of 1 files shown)


Razer6:
@jacobvosmaer Can you take a look?

sstanovnik:
The new format has "MD5:" prepended which parsed successfully but incorrectly, preventing users from using the SSH key, so I fixed that too.

jacobvosmaer:
Thanks for looking into this @sstanovnik

app/models/key.rb
```diff
@@ -83,15 +83,39 @@ def generate_fingerprint
...
+      if explicit_fingerprint_algorithm
+        cmd_output, cmd_status = popen(%W(ssh-keygen -E md5 -lf #{file.path}), '/tmp')
+      else
+        cmd_output, cmd_status = popen(%W(ssh-keygen -lf #{file.path}), '/tmp')
+      end
     end
 
     if cmd_status.zero?
-      cmd_output.gsub /(\h{2}:)+\h{2}/ do |match|
-        self.fingerprint = match
```
jacobvosmaer: This regex code looks a little wonky to me. I wonder if we can just pick out the MD5 fingerprint with something like:

```ruby
# constant should be defined somewhere at the top of the class
FINGERPRINT_REGEX = Regexp.new('\h{2}:' * 15 + '\h{2}' # 16 hex bytes separated by ':'
self.fingerprint = cmd_output.scan(FINGERPRINT_REGEX).first
```

jacobvosmaer: On the other hand, anchoring against the MD5 prefix seems like a nice defensive idea.

jacobvosmaer:
I like the basic idea but I am not 100% sure about the code style. What do you think @vsizov ?

vsizov:
@jacobvosmaer It looks good for me.

Razer6:
Already first issue reported related to this https://gitlab.com/gitlab-org/gitlab-ce/issues/1289

Razer6:
@vsizov @jacobvosmaer Any updates?

jacobvosmaer:
@Razer6: @DouweM will have a look at this.

DouweM:
I've cleaned the code up somewhat and posted a new MR to gitlab.com: https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/519

Thanks for your contribution @sstanovnik, I've credited you in the changelog.

cc @jacobvosmaer

kristianlm:
In case my quickfix is useful to anyone else:

```ruby
    Tempfile.open('gitlab_key_file') do |file|
      file.puts key
      file.rewind
      cmd_output, cmd_status = popen(%W(ssh-keygen -E md5 -lf #{file.path}), '/tmp')
    end

    if cmd_status.zero?
     puts "cmd_status was zero"
      cmd_output.match /MD5:(([\d\h]{2}:)+[\d\h]{2})/ do |match|
        self.fingerprint = match[1]
      end
    end
```

jacobvosmaer:
This got fixed in GitLab 7.10

PR #9012, opened 2015-03-24
Unescape branch param to delete

dantudor:
Branch names that contain `/` return a 405 error when being deleted because the slash is escaped to `%2F`
This patch will unescape the param prior to  executing the delete action.

Sorry, I am not a ruby developer and so am not able to create a test for this action.


randx:
Looks good. Thank you

PR #9023, opened 2015-03-25
Allow ability to delete branches with '/` in name

dantudor:
I initially thought the problem from my previous merge request was due to the URL escaping of the branch name.

Investigating the fault further and reading other methods in the commit it is actually a grape routing requirement that was missing from the delete method.


randx:
@dantudor I see tests are failing. Can you check?

tsigo:
Missing a comma before the requirements argument. You could probably fit them on the same line, as well.

dantudor:
@tsigo Thanks for fix. I broke the line to maintain the code style similar to the other methods in the class

PR #9034, opened 2015-03-26
Renamed Buildbox to Buildkite.

keithpitt:
Buildbox recently changed their name to Buildkite, so this PR updates the URL's and the text descriptions. It doesn't change the class name for backwards compatibility.


randx:
cool. Thank you

PR #9041, opened 2015-03-27
Add ActiveRecord::Migration.maintain_test_schema! to spec_helper

rspeicher:
New in Rails 4.1, this eliminates spec failures due to forgetting to run
`db:test:prepare`.


randx:
Thanks

PR #9048, opened 2015-03-29
properly paginate project events in API

### LINKED ISSUE #7492: Project events API not consistent with page & per_page parameters
The `GET /projects/:id/events` API doesn't work when it is invoked like the following,
`http://localhost/api/v3/projects/2/events?page=1&per_page=20` when there are less than 20 events. I get a `200 OK` but an empty json response. However it returns a good response when there are more than 20 events.

NARKOZ:
Fixes #7492


randx:
Thanks

PR #9056, opened 2015-03-30
Fix merge errors on CHANGELOG

mavimo:
During merges (d554070a and 497fd75d) changelog was "damaged", I restored the tagged 7.9.1 and added the required changes about 7.10.0 (unreleased). I think we need to ensure all PR are rebased on master before merge.


randx:
THank you!

PR #9058, opened 2015-03-30
upgrade hipchat

chulkilee:
In addition to several new features, it contains one bug fix for room name, which may affect gitlab too. See hipchat/hipchat-rb@31137b46cf12932bc914657b2c6176f149fad36d


jvanbaarsen:
@chulkilee Thanks

PR #9061, opened 2015-03-31
Ensure that the first added admin performs repository imports

fgimian:
A little change to ensure that the older administrator account is used for imports.  Typically this is the root account which often makes the most sense when performing an action from the CLI.

I noticed this problem after I imported some repositories and they were owned by our newest admin instead of root.

Thanks for your help!
Fotis

### DIFF
--- a/lib/tasks/gitlab/import.rake
+++ b/lib/tasks/gitlab/import.rake
@@ -35,7 +35,7 @@ namespace :gitlab do
         if project
           puts " * #{project.name} (#{repo_path}) exists"
         else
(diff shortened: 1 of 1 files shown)


jvanbaarsen:

fgimian:
Thanks, will do :smile:

randx:
Please add fix description to CHANGELOG

jvanbaarsen:
Closing this one in favour of: https://github.com/gitlabhq/gitlabhq/pull/9277

PR #7529, opened 2014-08-18
Users with group permission users should also create groups via API, fixes #6417

Razer6:
##### What does this MR do?

This PR determines ability to create a group, based on the `can_create_group` property of an user. This allows non-admins which are allowed to create groups, to also create them via API.
##### Why was this MR needed?

Currently only admins were allowed to do that.
##### What are the relevant issue numbers / Feature requests?

Fixes #6417

PR #9066, opened 2015-03-31
Fix #6417: users with group permission should be able to create groups via API

### LINKED ISSUE #6417: Unable to create groups by non-admin thru API
Even though I can create groups in GitLab itself (as I am granted with privilege to create groups) I cannot find a way how to force API to create group with current user as owner of the group and thus chance to add projects and members through API.

In other words: POST /groups should not be admin only, as not only admin has right to create group.

But following change (mad with respect to group_controller) doesn't cut it:

lib/api/groups.rb - line 43

```ruby
post do
  required_attributes! [:name, :path]

  attrs = attributes_for_keys [:name, :path]
  @group = Group.new(attrs)

  if @group.save
    @group.add_owner(current_user)
    present @group, with: Entities::Group
  else
    not_found!
  end
end
```

This creates group with name and path as intended, but current_user in not added to this group and whole presented result is a nil.

jirutka:
This is updated and slightly modified patch from #7529 that fixes #6417.

### DIFF
--- a/spec/requests/api/groups_spec.rb
+++ b/spec/requests/api/groups_spec.rb
@@ -3,8 +3,9 @@ require 'spec_helper'
 describe API::API, api: true  do
   include ApiHelpers
 
-  let(:user1) { create(:user) }
+  let(:user1) { create(:user, can_create_group: false) }
   let(:user2) { create(:user) }
+  let(:user3) { create(:user) }
   let(:admin) { create(:admin) }
   let!(:group1) { create(:group) }
   let!(:group2) { create(:group) }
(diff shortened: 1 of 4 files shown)


NARKOZ:
LGTM

deric:
+1

spec/requests/api/groups_spec.rb
```diff
@@ -3,8 +3,9 @@
 describe API::API, api: true  do
   include ApiHelpers
 
-  let(:user1) { create(:user) }
+  let(:user1) { create(:user, can_create_group: false) }
   let(:user2) { create(:user) }
+  let(:user3) { create(:user) }
```
jvanbaarsen: Is it needed to create a user3? Can't we just use user2?

jirutka: Yes, it’s needed. Read the code, you’ll see.

jvanbaarsen: Well, I see you use it. But in the diff I don;t see a reason you can't use user2. can you please explain in a little bit more wording why it is needed? I have about 100+ PR's that I need to review. I can't go and source dive all the time. Thats the reason I ask those questions. 
Its not my intention to insult you or anything like that. If you feel that way, I'm sorry.

I need to be sure that a PR is good before we can merge it. If I have the feeling it is not good I'm forced to close it without merging. That wouldn't help you, nor the GitLab project.

jirutka: We need a user that is not member of a group. Both user1 and user2 are set to be members of on of the groups.
I’ve send this PR one month ago, I really don’t remember all the details now…

jvanbaarsen: Ok, thanks for explaining :) :tada:

randx:
Thanks

PR #9074, opened 2015-04-01
Inform "per_page" and "page" default values

amaurymedeiros:
Inform the default values of the parameters "per_page" and "page". If this information is not explicit, the developers could lose a considerable amount of time wondering why a specific merge request is not being returned (specifically the 21st).


NARKOZ:
It's written in main README https://github.com/gitlabhq/gitlabhq/tree/master/doc/api#pagination

PR #9075, opened 2015-04-01
Bootlint integration

soullivaneuh:
As Gitlab theme is based on Twitter Bootstrap, it could be useful to have bootlint on dev env in order to fix and optimize bootstrap integration.

> Bootlint is a tool that checks for several common HTML mistakes in webpages that are using Bootstrap in a fairly "vanilla" way. Vanilla Bootstrap's components/widgets require their parts of the DOM to conform to certain structures. Bootlint checks that instances of Bootstrap components have correctly-structured HTML. Optimal usage of Bootstrap also requires that your pages include certain <meta> tags, an HTML5 doctype declaration, etc.; Bootlint checks that these are present.

Related repository: https://github.com/twbs/bootlint

Bootlint will produce error and warning on chrome debugger.

I disable alert notification to not be so annoyed while working onto another fix/feature.

What do you think?

Thanks.

### DIFF
--- a/app/views/layouts/_bootlint.haml
+++ b/app/views/layouts/_bootlint.haml
@@ -1,0 +1,4 @@
+:javascript
+  jQuery(document).ready(function() {
+    javascript:(function(){var s=document.createElement("script");s.onload=function(){bootlint.showLintReportForCurrentDocument([], {hasProblems: false, problemFree: false});};s.src="https://maxcdn.bootstrapcdn.com/bootlint/latest/bootlint.min.js";document.body.appendChild(s)})();
+  });
--- a/app/views/layouts/_head.html.haml
+++ b/app/views/layouts/_head.html.haml
@@ -16,6 +16,7 @@
 
   = render 'layouts/google_analytics' if extra_config.has_key?('google_analytics_id')
   = render 'layouts/piwik' if extra_config.has_key?('piwik_url') && extra_config.has_key?('piwik_site_id')
+  = render 'layouts/bootlint' if Rails.env == 'development'
 
   -# Atom feed
   - if current_user


jvanbaarsen:
@Soullivaneuh I like the idea, but since they also offer a bookmarklet (https://github.com/twbs/bootlint#in-the-browser) I would be more in favour of using that one. Otherwise we have to maintain that code snippet.

I'll close this PR for now, if you feel like this should be in the GitLab project, please let me know and I'll reconsider.

Soullivaneuh:
@jvanbaarsen 

> @Soullivaneuh I like the idea, but since they also offer a bookmarklet (https://github.com/twbs/bootlint#in-the-browser) I would be more in favour of using that one. Otherwise we have to maintain that code snippet.

I don't understand. My PR **is rightly** an implementation of this bookmarklet. Or I misunderstand your needs...

Regards

jvanbaarsen:
@Soullivaneuh Yeah I know its just an implementation of the bookmarklet. But as long as people use it in their browser using a bookmark, its their responsibility to update it. If we have it on the GitLab codebase, we need to maintain that bookmarklet.

Just to be sure, @randx What do you think?

Soullivaneuh:
@jvanbaarsen I understand your opinion. There is my thinking:

 * Implement it on gitlab dev will enforce contributors to respect bootstrap conventions (like houndci for Ruby). That's a must have.
 * The bootlint code integration is tiny and will not change very often. If it change and broke up, we just have to get the new code from bootlint repository, not a very big deal isn't it?

jvanbaarsen:
@Soullivaneuh I'll re-open and let @randx decide.

randx:
Lets try it

PR #9076, opened 2015-04-01
Fix event-title max-width in the dashboard

dsander:
For view port sizes between 1260px and 1200px pixels the event-title was
too wide, thus not truncated and got pushed into a second line.

100% - 174px was chosen as the max-width because 141px is the widest
.event-item-timestamp can get (with 'less than a minute ago') + 24px for
the avatar + 8px avatar margin + 1px to fit into the gap.

Before:
before

After:
after

For browsers that do not support `calc`, 70% width is the widest possible percentage value which prevents the title from wrapping into the second line.


randx:
:+1:

PR #9085, opened 2015-04-02
Remove slash add underscore

roshangautam:
Using NAMESPACE/PROJECT_NAME causes problems in apache and needs <code>AllowEncodedSlashes NoDecode</code> parameter enabled in vhost configuration file of apache. In worst case sometimes just enabling this parameter doesn't work.

To avoid this situation i used a different separator  _ (underscore) to make things for simple

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -224,9 +224,9 @@ class Project < ActiveRecord::Base
     end
 
     def find_with_namespace(id)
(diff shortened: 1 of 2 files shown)


roshangautam:
Instead of using _ it can also be fixed by making sure that the id is decoded before using it. 

<code>id = URI.decode(id)</code>

in <code>find_with_namespace</code>

PR #9094, opened 2015-04-06
update text to reference new location of setting

bbodenmiller:
Setting for Restricted visibility levels is now under Admin area.


jvanbaarsen:
@bbodenmiller Thanks!

PR #9100, opened 2015-04-08
improve new project options

bbodenmiller:
- Simplify new repo commands
- Include directions for how to add an existing folder that is either already a repo or not yet a repo

Replaces #7918.

Before:
image

After:
image

[No harm in initing an already inited repo](https://git-scm.herokuapp.com/docs/git-init):

> Running git init in an existing repository is safe. It will not overwrite things that are already there.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -10,6 +10,7 @@ v 7.11.0 (unreleased)
   -
   -
   -
+  - Improve new project command options (Ben Bodenmiller)
 
 v 7.10.0 (unreleased)
   - Ignore submodules that are defined in .gitmodules but are checked in as directories.
--- a/app/views/projects/empty.html.haml
+++ b/app/views/projects/empty.html.haml
@@ -27,20 +27,19 @@
     %legend Create a new repository
     %pre.dark
       :preserve
-        mkdir #{@project.path}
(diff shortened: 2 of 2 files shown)


jvanbaarsen:
@bbodenmiller Do we need to touch README.md and add it to git of the folder already exists? We don't want to have users think they need to have that there.

I'm not really in favour of this change since for me it does not add any clarification. @randx What do you think?

bbodenmiller:
@jvanbaarsen adding readme isn't needed in either bucket, it just demos how to add a file & make a commit. I'm open to other ideas. My main objective here is to document how a user with a bunch of files in a non-git folder can turn that in to a Git repo and start commiting files. The current directions assume you are starting from scratch or have an existing repo.

jvanbaarsen:
@bbodenmiller Ok, what about just adding the `git init` command?

bbodenmiller:
Works for me. I'll update later.

bbodenmiller:
@jvanbaarsen okay, updated. I don't have updated screenshots but you can see the general idea in the code.

jvanbaarsen:
@bbodenmiller Thanks

randx:
:+1:

PR #9114, opened 2015-04-10
Include creator_id in project info of API

dreis2211:
Hey guys,

I recently crawled over our Gitlab instance by using the API to find empty repositories to contact their creators to either fill them with live or to delete them. The owner wasn't always available as the projects were created inside a group, so I had no good information about the real creator.

I used one of the owners of the namespace to workaround it for now, but I propose to add the creator_id to the project info of the public API.

Since we're not deleting users from the tool, but only disable them this could also be sufficient to find out if a project's creator is not available anymore by simply getting the user through the API and fetching its status. If he was the only contributor, a project might get lost, because it won't be maintained anymore. With that little info you could eventually find new main contributor(s) for a project - or at least you would know that the initial creator isn't available anymore.

Cheers

### DIFF
--- a/doc/api/projects.md
+++ b/doc/api/projects.md
@@ -63,6 +63,7 @@ Parameters:
     "snippets_enabled": false,
     "created_at": "2013-09-30T13: 46: 02Z",
     "last_activity_at": "2013-09-30T13: 46: 02Z",
+    "creator_id": 3,
     "namespace": {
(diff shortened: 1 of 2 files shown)


randx:
Can you please update API documentation as well?

dreis2211:
Done

randx:
@dreis2211 thank you!

PR #9122, opened 2015-04-13
Export users to CSV file

bbodenmiller:
This PR adds the ability for a system admin to export a list of users and basic account details via a rake task. The output will be written to `users.csv` in the configured backup folder.

```
# use this command if you've installed GitLab with the Omnibus package
sudo gitlab-rake gitlab:backup:users

# if you've installed GitLab from source
sudo -u git -H bundle exec rake gitlab:backup:users RAILS_ENV=production
```

The motive for adding this feature is so that system administrators can keep a list of users & emails separate from the GitLab installation in case of prolonged downtime. Additionally this allows admins to quickly verify some account details (project limit, can create group, etc.) not yet easily bulk verified in the UI.

As this is related to backups it was added to backup rake file. At this time the generated file is not included in the backup tar file nor generated when `gitlab:backup:create` is run. In the future it may be valuable stored in the backup tar file however at this time the folks I have spoke with see more value in having `users.csv` generated and stored separately.

### DIFF
--- a/lib/backup/users.rb
+++ b/lib/backup/users.rb
@@ -1,0 +1,18 @@
+module Backup
+  class Users
+    attr_reader :user_dump_file
+
+    def initialize
+      @user_dump_file = File.join(Gitlab.config.backup.path, 'users.csv')
+    end
+
+    def dump
+      CSV.open(user_dump_file, "w") do |csv|
+        csv << ["email", "name", "admin", "projects_limit", "username", "can_create_group", "state"]
+        User.all.each do |u|
+          csv << [u.email, u.name, u.admin, u.projects_limit, u.username, u.can_create_group, u.state]
+        end
+      end
+    end
+  end
+end
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -73,6 +73,7 @@ v 7.10.0 (unreleased)
   - Fix git over ssh errors 'fatal: protocol error: bad line length character' 
   - Automatically setup GitLab CI project for forks if origin project has GitLab CI enabled
   - Bust group page project list cache when namespace name or path changes.
+  - Add backup users list to CSV file option (Ben Bodenmiller)
 
 v 7.9.3
   - Contains no changes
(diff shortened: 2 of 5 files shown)


bbodenmiller:
@Razer6 @jacobvosmaer who should I ping for a review?

lib/backup/users.rb
```diff
@@ -0,0 +1,18 @@
...
+    attr_reader :user_dump_file
+
+    def initialize
+      @user_dump_file = File.join(Gitlab.config.backup.path, 'users.csv')
+    end
+
+    def dump
+      CSV.open(user_dump_file, "w") do |csv|
+        csv << ["email", "name", "admin", "projects_limit", "username", "can_create_group", "state"]
+        User.all.each do |u|
```
DouweM: For more efficient memory usage, use `User.find_each` instead of `User.all.each`

DouweM:
I like the functionality, but I think it may be confusing to have it under the backup namespace when it doesn't actually run as part of the backup task. Not a really big deal though.

@jacobvosmaer What do you think?

jacobvosmaer:
I agree with @DouweM that this does not really belong in the 'backup' category; there is no way to restore this information. Maybe `rake gitlab:export:users`?

I also think it would be better to just write the CSV to standard output. There is no technical need for the rake task to create a file (is there?), it makes the code simpler, and you can do fun stuff like piping it into grep.

lib/backup/users.rb
jacobvosmaer: :+1: for find_each

bbodenmiller:
@jacobvosmaer @DouweM all good points. Only thing I'm unsure of is how to write some good tests for this.

jacobvosmaer:
@bbodenmiller maybe create a function that does all the work and takes an IO object as an argument. Then the rake task would do `myfunc($stdout)` and your test would to `result = StringIO.new; myfunc(result); expect(result) { bla} `.

jvanbaarsen:

PR #9129, opened 2015-04-13
Update to latest gemojione version

jonathanwiesel:
Latest version adds missing `:memo:` - :memo: emoji


Razer6:
This is the wrong place to pinpoint the version. The right place to update the dependency is here: https://gitlab.com/gitlab-org/gitlab_emoji

PR #9150, opened 2015-04-17
do not prepend relative_url_root to avatar

costela:
The commit.author.icon variable contains a full URL, which gets broken
when prepended by the relative_url_root setting (generating hrefs like
"/gitlabhttps://...".


jvanbaarsen:

costela:
It was caused by an unrelated problem. The current build passed. Thanks for the heads-up!

jvanbaarsen:
@costela Thanks

PR #9164, opened 2015-04-21
Make issue preview look like the result

dsander:
Adding the `md` class to the `md-preview-holder` div applies the styles of the markdown formatting `md-typography` mixin. Which makes the preview look like the actual result.

Before: 
before
After:
after


jvanbaarsen:
@dsander Thanks!

PR #9185, opened 2015-04-24
API: recursive tree request

rebecamendez:
As i've suggested in http://feedback.gitlab.com/forums/176466-general/suggestions/6661459-api-allow-recursive-repository-tree: added a parameter in tree request to get a recursive tree and path attribute in response.

### DIFF
--- a/app/models/repository.rb
+++ b/app/models/repository.rb
@@ -220,7 +220,7 @@ class Repository
     @head_tree ||= Tree.new(self, head_commit.sha, nil)
   end
 
-  def tree(sha = :head, path = nil)
+  def tree(sha = :head, path = nil, recursive = false)
     if sha == :head
       if path.nil?
         return head_tree
@@ -229,7 +229,7 @@ class Repository
       end
     end
 
-    Tree.new(self, sha, path)
+    Tree.new(self, sha, path, recursive)
   end
 
   def blob_at_branch(branch_name, path)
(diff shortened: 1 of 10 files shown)


app/models/repository.rb
```diff
@@ -220,7 +220,7 @@ def head_tree
     @head_tree ||= Tree.new(self, head_commit.sha, nil)
   end
 
-  def tree(sha = :head, path = nil)
+  def tree(sha = :head, path = nil, recursive = 0)
```
jvanbaarsen: Why did you choose `0` instead of false?

app/models/tree.rb
```diff
@@ -3,15 +3,26 @@ class Tree
 
   attr_accessor :repository, :sha, :path, :entries
 
-  def initialize(repository, sha, path = '/')
+  def initialize(repository, sha, path = '/', recursive = '0')
```
jvanbaarsen: Is there a reason to be using `'0'` over `false`?

rebecamendez:
I use the Github syntax: https://developer.github.com/v3/git/trees/#get-a-tree-recursively, but if you prefer we can change the flag to a boolean value.

jvanbaarsen:
@rebecamendez Hm ok, but maybe we can translate the URL params to actual booleans where they come in? That way we don't need to pass around `1` and `0`.

rebecamendez:
I made a mistake with the rebase, so I start a new pull request: https://github.com/gitlabhq/gitlabhq/pull/9311

PR #9189, opened 2015-04-24
Removed "unreleased" from v7.10 changelog

rockneverdies55:
Version 7.10 is already released so "unreleased" text next to it in the changelog file can be removed.


Razer6:
Thanks

PR #9196, opened 2015-04-25
fix redis 3.0.0

tonicmuroq:
when I use redis-3.0, `sudo -u git -H bundle exec rake gitlab:check RAILS_ENV=production` can't pass.

### DIFF
--- a/lib/tasks/gitlab/check.rake
+++ b/lib/tasks/gitlab/check.rake
@@ -282,7 +282,8 @@ namespace :gitlab do
     def check_redis_version
       print "Redis version >= 2.0.0? ... "
 
-      if run_and_match(%W(redis-cli --version), /redis-cli 2.\d.\d/)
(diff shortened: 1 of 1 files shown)


lib/tasks/gitlab/check.rake
```diff
@@ -282,7 +282,8 @@ namespace :gitlab do
     def check_redis_version
       print "Redis version >= 2.0.0? ... "
 
-      if run_and_match(%W(redis-cli --version), /redis-cli 2.\d.\d/)
+      redis_version = run(%W(redis-cli --version))
+      if redis_version.try(:match, /redis-cli 2.\d.\d/) or redis_version.try(:match, /redis-cli 3.\d.\d/)
```
jvanbaarsen: Can you please use the `||` or variant?

tonicbupt:
@jvanbaarsen I change it to `||` and rebased commits :3

PR #9200, opened 2015-04-25
Version 7.10.0 has been released so its 7.11.0-pre time

jvanbaarsen:
Version 7.10.0 has been released, so I think the *-pre should be dropped.


maxlazio:
@jvanbaarsen Since 7.10 has been released we should be ad `7.11.0.pre` :)

jvanbaarsen:
@maxlazio Ah ok :D Almost good :P Will fix tomorrow :)

jvanbaarsen:
@maxlazio Fixed.

Razer6:
Thank you

PR #9207, opened 2015-04-26
Fix missing X-UA-Compatible meta tag

soullivaneuh:
Fixes bootlint [W002](https://github.com/twbs/bootlint/wiki/W002) issue.


jvanbaarsen:
Thanks

PR #9212, opened 2015-04-27
Project deletion performance fix

KelvinMutuma:
@randx Fix for [issue #1163](https://gitlab.com/gitlab-org/gitlab-ce/issues/1163). i have used delete_all instead of destroy_all to generate a single query for deletion of a project and its related data to optimise on speed


KelvinMutuma:
Closed this due to failed build

PR #9220, opened 2015-04-28
Fix #9219 Affix menu size change

### LINKED ISSUE #9219: Affix menu size change
Hi, the affix menu has not the same size when it is on the top and when it go down since it is a fixed with (220px) I propose to use margin instead of width to avoid that behaviour.

|                      | Scroll to the top | Scroll to the bottom |
|------------------|---------------------|--------------------------|
|width: 220px;  | before-affix-top | before-affix-middle |
| margin-right: 35px; | before-affix-top | after-affix-middle |

kylekatarnls:
Before / after the change

|  | Scroll to the top | Scroll to the bottom |
| --- | --- | --- |
| width: 220px; | before-affix-top | before-affix-middle |
| margin-right: 35px; | before-affix-top | after-affix-middle |

### DIFF
--- a/app/assets/stylesheets/pages/issuable.scss
+++ b/app/assets/stylesheets/pages/issuable.scss
@@ -19,7 +19,7 @@
     &.affix {
       position: fixed;
       top: 70px;
-      width: 220px;
+      margin-right: 35px;
     }
   }
 }


jvanbaarsen:
@kylekatarnls Can you please post a before and after screenshot of this change?

kylekatarnls:
Here is the screenshots. In fact the menu is reduced when I scroll.

PR #9208, opened 2015-04-26
Fix .container position on navbar

soullivaneuh:
Fixes bootlint [W012](https://github.com/twbs/bootlint/wiki/W012) issue.

PR #9223, opened 2015-04-29
Fix .container position on navbar (reloaded)

soullivaneuh:
This is a fixed on improved version of #9208.

I simply remove `navbar-inner` that is not needed anymore.

I also apply changes on `_empty_head_panel` and `_head_panel` templates.

I tested it with both desktop and mobile view on dashboard main page, no apparent side effect.

@jvanbaarsen @randx @Razer6 => Ready for review. :+1:

### DIFF
--- a/app/assets/stylesheets/generic/header.scss
+++ b/app/assets/stylesheets/generic/header.scss
@@ -10,7 +10,10 @@ header {
     border: none;
     width: 100%;
 
-    .navbar-inner {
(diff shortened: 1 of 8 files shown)


Soullivaneuh:
Crap. I did something wrong on commit. Let me fix that.

Soullivaneuh:
Fixed in #9224. Sorry for that.

PR #9224, opened 2015-04-29
Fix .container position on navbar (reloaded)

soullivaneuh:
This is a fixed and improved version of #9208.

I simply remove `navbar-inner` that is not needed anymore.

I also apply changes on `_empty_head_panel` and `_head_panel` templates.

I tested it with both desktop and mobile view on dashboard main page, no apparent side effect.

@jvanbaarsen @randx @Razer6 => Ready for review. :+1:


randx:
Thanks!

PR #9233, opened 2015-04-30
Add commit streak to user profile page

goorockey:
Add commit streak information like Github to user profile page.

Before
before

After
after

Relevant feature request: 
http://feedback.gitlab.com/forums/176466-general/suggestions/5863108-implement-github-like-commit-streak

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -50,6 +50,10 @@
 #  bitbucket_access_token        :string(255)
 #  bitbucket_access_token_secret :string(255)
 #  public_email                  :string(255)      default(""), not null
+#  longest_streak_start_at       :date
+#  longest_streak_end_at         :date
+#  current_streak                :integer          default(0)
+#  last_contributed_at           :date
 #
 
 require 'carrierwave/orm/activerecord'
@@ -148,7 +152,7 @@ class User < ActiveRecord::Base
 
   before_save :ensure_authentication_token
   after_save :ensure_namespace_correct
-  after_initialize :set_projects_limit
+  after_initialize :set_projects_limit, :set_current_streak
   after_create :post_create_hook
   after_destroy :post_destroy_hook
 
@@ -459,6 +463,14 @@ class User < ActiveRecord::Base
     self.projects_limit = current_application_settings.default_projects_limit
   end
 
+  def set_current_streak
+    if last_contributed_at.nil? || last_contributed_at + 1 < Date.today
+      self.current_streak = 0
+    end
+
+    true
+  end
+
   def requires_ldap_check?
     if !Gitlab.config.ldap.enabled
       false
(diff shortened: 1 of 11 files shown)


jvanbaarsen:
@randx Would you be interested in something like this?

rumpelsepp:
Personally, just because it's cool. :D

randx:
The idea is cool. But only if it will not affect performance

jvanbaarsen:
@goorockey I'm missing overall tests for this feature. Can you add those? Also can you check if this would impact performance?

app/models/user.rb
```diff
@@ -611,4 +626,63 @@ def contributed_projects_ids
...
+    end
+
+    if self.current_streak == 1
+      return streak_date_format(self.last_contributed_at)
+    end
+
+    streak_date_format(self.last_contributed_at - self.current_streak + 1) + ' - ' + streak_date_format(self.last_contributed_at)
+  end
+
+  def streak
```
jvanbaarsen: can you drop the `self.` if it is not needed?

app/models/user.rb
```diff
@@ -611,4 +626,63 @@ def contributed_projects_ids
...
+    if self.last_contributed_at == Date.today
```
jvanbaarsen: can you write this as follow:
```
return if last_contributed_at == Date.Today
``

File: app/models/user.rb [hunk 3]
```diff
@@ -611,4 +626,63 @@ def contributed_projects_ids
...
+      return true
+    end
+
+    if self.last_contributed_at != nil && self.last_contributed_at + 1 == Date.today
```
jvanbaarsen: Maybe you can write this as:

```
current_streak ||= 1
current_streak +=1 if last_contributed_at && last_contributed_at + 1 == Date.today


```

app/models/user.rb
```diff
@@ -611,4 +626,63 @@ def contributed_projects_ids
...
+      self.current_streak += 1
+    else
+      self.current_streak = 1
+    end
+
+    if self.current_streak > self.longest_streak
+      self.longest_streak = self.current_streak
```
jvanbaarsen: Do we need to save the longest_streak? we can also calculate it since we have the start and the end date right? What do you think about that approach?

app/models/user.rb
```diff
@@ -140,6 +145,8 @@ class User < ActiveRecord::Base
   validate :unique_email, if: ->(user) { user.email_changed? }
   validate :owns_notification_email, if: ->(user) { user.notification_email_changed? }
   validates :avatar, file_size: { maximum: 200.kilobytes.to_i }
+  validates :longest_streak, numericality: { greater_than_or_equal_to: 0 }
```
jvanbaarsen: Since we don't store this information by a user action, I think we can drop these 2 validations.

app/models/user.rb
```diff
@@ -459,6 +466,14 @@ def set_projects_limit
     self.projects_limit = current_application_settings.default_projects_limit
   end
 
+  def set_current_streak
```
jvanbaarsen: Do we really need to do this? What about setting a default value in the database of `0`?

app/models/user.rb
```diff
@@ -611,4 +626,63 @@ def contributed_projects_ids
       select(:project_id).
       uniq.map(&:project_id)
   end
+
+  def contribution_year_period
+    streak_date_format(1.year.ago) + ' - ' + streak_date_format(Date.today)
+  end
+
+  def longest_streak_period
+    if self.longest_streak == 0
```
jvanbaarsen: Can you write this as:

```
return 'No contribution yet' if longest_streak.zero?
```

app/models/user.rb
```diff
@@ -611,4 +626,63 @@ def contributed_projects_ids
...
+
+    if self.longest_streak_start_at == self.longest_streak_end_at
+      return streak_date_format(self.longest_streak_start_at)
+    end
+
+    streak_date_format(self.longest_streak_start_at) + ' - ' + streak_date_format(self.longest_streak_end_at)
+  end
+
+  def current_streak_period
+    if self.longest_streak == 0
```
jvanbaarsen: Can you write this as:

return 'No contribution yet' if longest_streak.zero?

app/models/user.rb
```diff
@@ -611,4 +626,63 @@ def contributed_projects_ids
...
+      return 'No contribution yet'
+    end
+
+    if self.last_contributed_at + 1 < Date.today
+      return 'Last contributed at ' + streak_date_format(self.last_contributed_at)
+    end
+
+    if self.current_streak == 1
+      return streak_date_format(self.last_contributed_at)
```
jvanbaarsen: Can you write this as:

```ruby
return streak_date_format(last_contributed_at) if current_streak == 1
```

app/models/user.rb
goorockey: We need this to reset current_streak to zero automatically when the last_contributed day is pass by.

app/models/user.rb
goorockey: Code is refactored at 8c96cca. :D

goorockey:
Yes, I'd love to add some tests and check out the performance.

app/models/user.rb
jvanbaarsen: But is that something that has to do with initializing a new user? Maybe we can make this more explicit?

app/models/user.rb
```diff
@@ -611,4 +623,57 @@ def contributed_projects_ids
...
+      streak_date_format(last_contributed_at)
+    else
+      streak_date_format(last_contributed_at - current_streak + 1) + ' - ' + streak_date_format(last_contributed_at)
+    end
+  end
+
+  def streak
+    return if last_contributed_at == Date.today
+
+    self.current_streak ||= 1
```
jvanbaarsen: Do we need the `self` here?

jvanbaarsen:
@goorockey Are you still working on this?

goorockey:
@jvanbaarsen Yes. But I'm too busy on my own project, and maybe cannot work on this until the day after tomorrow. I'm very sorry.

jvanbaarsen:
@goorockey I'll close this until you have time to work on this again, ok? Ping me when you want me to re-open.

PR #9234, opened 2015-04-30
Group milestones by title in the dashboard and all other issue views

dsander:
This groups milestones by title for issue views like it has been done for the milestone dashboard/project overview. Before milestones with the same title would show up multiple times in the filter dropdown and one could only filter per project and milestone. Now the milestone filter is based on the title of the milestone, i.e. all issues marked with the same milestone title are shown.

While working on this I noticed that the variables assigned by `ApplicationController#set_filter_values` seem not to be used by anything anymore, so I removed the method and it's invocations. I included it in this PR because I would have to make changes to it (`milestone_id` vs `milestone_title`) whose consequences I could not test. If I missed something, please correct me, and I will revert the second commit and change it to use the `milestone_label`.

~~The specs do not work very well locally for me, so I am labeling it as WIP until the CI confirms the specs pass.~~

Before:
before
After:
after

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -20,6 +20,7 @@ v 7.11.0 (unreleased)
   - Don't crash when an MR from a fork has a cross-reference comment from the target project on of its commits.
   - Include commit comments in MR from a forked project.
   - Fix adding new group members from admin area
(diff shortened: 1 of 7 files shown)


randx:
Great! Thank you

PR #9240, opened 2015-05-04
Fix branch labels being underlined as links

awiddersheim:
When hovering over the branch labels they are underlined since they are
part of the link to the branch. This does not look very good. Removed
branch labels as being part of the link to the branch.

This could also be fixed with CSS of course but decided to go with this quick
fix instead.

Here is a screen shot of the problem:

gitlab

### DIFF
--- a/app/views/projects/branches/_branch.html.haml
+++ b/app/views/projects/branches/_branch.html.haml
@@ -1,14 +1,13 @@
 - commit = @repository.commit(branch.target)
 %li(class="js-branch-#{branch.name}")
   %h4
-    = link_to namespace_project_tree_path(@project.namespace, @project, branch.name) do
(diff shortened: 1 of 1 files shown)


awiddersheim:
This doesn't actually fix the problem very well. You end up with this when using the code in this pull request:

image

Note the lack of space between the branch and labels. This issue is also present when viewing the branches on the `settings -> protected branches` page as well. Toyed with a number of options but didn't really find a great solution unfortunately. Leaving pull request open as it hopefully serves as a good issue report even if the code isn't very good.

jvanbaarsen:
@awiddersheim Can you maybe open up an issue for this? I'll close this PR since the code is not an improvement on the current implementation. Thanks for the efforts though!

PR #9252, opened 2015-05-06
List of default web hooks could be setted up

newmen:
Now we can setup our GitLab for auto create web hooks for each new project

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -50,18 +50,17 @@ class Project < ActiveRecord::Base
   default_value_for :wall_enabled, false
   default_value_for :snippets_enabled, gitlab_config_features.snippets
 
-  # set last_activity_at to the same as created_at
-  after_create :set_last_activity_at
-  def set_last_activity_at
-    update_column(:last_activity_at, self.created_at)
-  end
-
-  # create default hooks if they are presented
   after_create :create_default_web_hooks, if: -> { !!gitlab_config_features.auto_web_hooks }
   def create_default_web_hooks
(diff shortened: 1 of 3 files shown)


config/gitlab.yml.example
```diff
@@ -80,6 +80,11 @@ production: &base
       wiki: true
       snippets: false
 
+      ## Default web hooks settings
```
jvanbaarsen: @randx We are trying to move all the settings to the settings page right? So no new settings should be added to gitlab.yml?

app/models/project.rb
```diff
@@ -56,6 +56,14 @@ def set_last_activity_at
     update_column(:last_activity_at, self.created_at)
   end
 
+  # create default hooks if they are presented
+  after_create :create_default_web_hooks, if: -> { !!gitlab_config_features.auto_web_hooks }
```
jvanbaarsen: Can you please move the after_create to the top of the file? That way we can see in one look what callbacks are being called. Also the comment seems to be redundant, you can remove that one

config/gitlab.yml.example
randx: @jvanbaarsen right. Except cases like path to `git` binary

jvanbaarsen: Ok, so for this PR the config needs to be moved to some web interface right? Can you say if you would like to see this merged in? If so I'll guide the rest of the PR.

dblessing:
@newmen Can you please change this to put the setting in the admin section of the UI?

jvanbaarsen:

PR #9276, opened 2015-05-10
Extract handling of relative file links to its own HTML filter

### LINKED ISSUE #9263: Relative links doesn’t work in AsciiDoc files
When I use a relative link inside an `.adoc` file, e.g. `link:LICENSE` in `README.adoc`, then the rendered link is incorrect, `/{username}/LICENSE` instead of `/{username}/{project-name}/blob/master/LICENSE`.

This bothers the most in the case of images. My colleagues currently works on some technical reports and they’re displaying these primarily on GitLab and they are not much happy for not seeing their graphs etc.

jirutka:
I’ve extracted rebuilding of relative links from GitlabMarkdownHelper to its own HTML filter (RelativeLinkFilter) as suggested in comment [here](https://github.com/gitlabhq/gitlabhq/blob/e46d1cdd8bd4cc12e8c8e8fdce10b3114a17d95e/app/helpers/gitlab_markdown_helper.rb#L75).

My motivation for this is to fix #9263 and this is the first step. HTML filters can be easily reused for both Markdown and AsciiDoc.

However, I’m not quite sure about this, the code around Markdown rendering is very… complicated. @randx, could you please review it and give me some feedback?


tsigo:
Awesome! I'll review in the morning.

lib/gitlab/markdown.rb
```diff
@@ -91,6 +97,8 @@ def filters
       [
         Gitlab::Markdown::SanitizationFilter,
 
+        Gitlab::Markdown::RelativeLinkFilter,
+
```
tsigo: You can remove this newline and have it grouped with the Emoji/ToC/Autolink filters.

lib/gitlab/markdown/relative_link_filter.rb
```diff
@@ -0,0 +1,117 @@
+require 'uri'
```
tsigo: Add `require 'html/pipeline/filter'` before this line.

lib/gitlab/markdown/relative_link_filter.rb
```diff
@@ -0,0 +1,117 @@
...
+      def rebuild_relative_uri(uri)
+        file_path = relative_file_path(uri.path)
+
+        uri.path = [
+          relative_url_root,
+          project.path_with_namespace,
+          path_type(file_path),
+          ref || 'master',  # assume that if no ref exists we can point to master
+          file_path
+        ].compact.join('/').squeeze('/').chomp('/')
```
tsigo: After `compact` removes nil values, is it even possible there would be duplicate slashes requiring `squeeze`, or a trailing slash requiring `chomp`?

lib/gitlab/markdown/relative_link_filter.rb
```diff
@@ -0,0 +1,117 @@
...
+          repository.head_commit.sha
+        end
+      end
+
+      def relative_url_root
+        Gitlab.config.gitlab.relative_url_root.presence || '/'
+      end
+
+      [:commit, :project, :project_wiki, :requested_path, :ref].each do |name|
+        define_method(name) do
```
tsigo: I understand wanting to reduce repetitive, boilerplate code, but I'd prefer these to each be their own standard method, just to be more explicit.

tsigo:
:+1: Thanks for this!

I think the only specs we have covering this right now are those in `spec/helpers/gitlab_markdown_helper_spec.rb`. Would you be able to convert those into unit tests for your new filter?

lib/gitlab/markdown/relative_link_filter.rb
```diff
@@ -0,0 +1,117 @@
...
+    # Context options:
+    #   :commit
+    #   :project
+    #   :project_wiki
+    #   :requested_path
+    #   :ref
+    class RelativeLinkFilter < HTML::Pipeline::Filter
+
+      def call
+        if !project_wiki && repository.try(:exists?) && !repository.empty?
```
tsigo: Maybe extract this group of conditions into a helper method, like `relative_linkable?` to aid in testing.

lib/gitlab/markdown/relative_link_filter.rb
jirutka: Done.

lib/gitlab/markdown.rb
```diff
@@ -64,7 +65,12 @@ def gfm_with_options(text, options = {}, project = @project, html_options = {})
         current_user:    current_user,
         only_path:       options[:reference_only_path],
         project:         project,
-        reference_class: html_options[:class]
+        reference_class: html_options[:class],
+
+        # RelativeLinkFilter
+        ref: @ref,
+        requested_path: @path,
+        project_wiki: @project_wiki
```
jirutka: @tsigo I have bad feeling from depending on instance variables here. What do you think? Is it safe to assume that these variables are always present?

tsigo: If they were available to the old helper, they should be available here. If all three are truly required by the filter, you can check at the beginning of `call` and `return doc` if any of them are nil.

jirutka:
@tsigo I’ve added spec for RelativeLinkFilter. It’s not perfect, but I don’t have time to cover all possible cases. Hover, it already covers more cases than the previous one.

tsigo:
@jirutka Ok. I've got some minor stylistic nitpicks with the spec, but I can fix those myself in a future MR. Thank you again for taking this on.

@randx This should be good to merge.

jirutka:
@randx CI finally works and all tests passes, so :shipit:? :smiley_cat:

randx:
:+1:

tsigo:
@jirutka Could you please look into https://gitlab.com/gitlab-org/gitlab-ce/issues/1596

jirutka:
@tsigo I look at it, but I don’t understand the issue. Some example would be helpful.

tsigo:
@jirutka I believe the issue is that the project's default branch is set to something other than master, but when no ref is given we still default to master in the filter [here](https://github.com/gitlabhq/gitlabhq/pull/9276/files#diff-dda715e9933e814a6e436a0521f942b9R52). Not 100% sure.

jirutka:
@tsigo Ah, I see. However, I haven’t created this assumption, it’s taken from the previous code, see [here](https://github.com/gitlabhq/gitlabhq/blob/e46d1cdd8bd4cc12e8c8e8fdce10b3114a17d95e/app/helpers/gitlab_markdown_helper.rb#L210-L213).

jirutka:
@tsigo Maybe we should replace `ref || 'master'` with `ref || context[:project].default_branch`?

PR #8800, opened 2015-02-15
Allow to configure smtp and sendmail in gitlab.yml

jirutka:
Since almost all GitLab settings are configured in gitlab.yml, it’s very inconvenient to configure SMTP server in `config/initializers/smtp_settings.rb`.

I’ve named it explicitly `outgoing_emails`, because I assume that you may someday implement an ability to respond to emails from GitLab and then you’ll need to distinguish these settings.

PR #9281, opened 2015-05-11
Allow to configure smtp and sendmail in gitlab.yml

jirutka:
Reopening #8800.

TODO:
- [ ] Make it backward compatible with existing `smtp_settings.rb` (see comments [here](https://github.com/gitlabhq/gitlabhq/pull/8800#issuecomment-99359016) and [here](https://github.com/gitlabhq/gitlabhq/pull/8800#issuecomment-99494266)).
- [ ] Create `doc/update/7.10-to-7.11.md` and add instructions to migrate email settings from `smtp_settings.rb` to `gitlab.yml`.
- [ ] Mention this change in the changelog.
- [ ] Update Omnibus package → [omnibus-gitlab#555](https://gitlab.com/gitlab-org/omnibus-gitlab/issues/555).


config/gitlab.yml.example
```diff
@@ -89,6 +79,57 @@ production: &base
...
+    smtp_settings:
+      # Hostname of the SMTP server used to send emails (default: localhost).
+      address: localhost
+
+      # Port of the SMTP server used to send emails (default: 25).
+      port: 25
+
+      # If your mail server requires authentication, you need to specify the
+      # authentication type here. Possible values are:
+      #   :plain    - send the password in the clear text.
```
axilleas: The `plain` method does not send the password in cleartext. See https://github.com/rails/rails/blob/master/actionmailer/lib/action_mailer/base.rb#L383 and https://github.com/rails/rails/pull/19930

dserodio: base64 _is_ cleartext. It's not encrypted or hashed

axilleas: I see your point and I agree, but we have two methods that behave almost the same, `:plain` and `:login` and they are both base64 encoded. The way it is written now, one might think `:login` is stronger than `:plain`.

I don't know how you define cleartext, but in my mind it means that a password is read as is. However in base64 you'd have to do a decode first, although trivial.

jirutka: What’s difference between `plain` and `login`?

jvanbaarsen:

PR #9291, opened 2015-05-13
Fixed issue where email address was missing from API calls.

mransley:
The email address attribute was missing from group and project memberships, this used to appear on a previous release and this pull request is just adding the attribute back in.

### DIFF
--- a/lib/api/entities.rb
+++ b/lib/api/entities.rb
@@ -5,7 +5,7 @@ module API
     end
 
     class UserBasic < UserSafe
-      expose :id, :state, :avatar_url
+      expose :id, :state, :avatar_url, :email
     end
 
     class User < UserBasic


jvanbaarsen:
@mransley This has been removed on purpose. See https://github.com/gitlabhq/gitlabhq/commit/ae564c97d48bf728745c57720734cb40378fd90f

I'll close this PR.

mransley:
Agreed that it shouldn't be leaked out for a standard user, but this was an admin user performing the call - should the merge be changed to return the email address if the user performing the API call is an admin user?

jvanbaarsen:
@mransley When you're an admin you get the full user information back: https://github.com/gitlabhq/gitlabhq/commit/ae564c97d48bf728745c57720734cb40378fd90f#diff-60dced23f72e147e2b1489226eea06d3R17

mransley:
You don't get the email address when you get the membership of a group or project (which you used to be able to).  The only way to get the email address is to perform another API call to get the user details - which for my client is going to lead to a significant number of additional API calls for what they are trying to achieve.

jvanbaarsen:
@randx Do you see problems with making the email public (for admins) when you get memberships of a group or project?

randx:
I am ok with admin having access to user emails via API. But in this PR it just expose it to every place where `UserBasic` is used

jvanbaarsen:
@randx Yeah I know, thats why I closed it. @mransley If you want, you can create a new PR that only exposes the email to the admin. When you create a new PR, can you mention this PR in there?

PR #8768, opened 2015-02-10
Add Composer Project Service (updated)

ikappas:
This commit adds the per-project option to expose a project as a composer package. It generates and maintains a packages.json file along with a project specific /p/project-{id}.json file (conforming to https://getcomposer.org/doc/05-repositories.md#includes) in the public/ folder, allowing the GitLab install to be used as a Composer repository.

```
Note: There is a known issue integrating with the GitLab-Omnibus, namely that the public folder is owned by 'root' and GitLab is run as 'git'. The GitLab-Omnibus chef script needs to be updated to allow access to both packages.json and the /p/ folder in the public/ folder
```

It extends on the original idea providing extended service configuration options that allow to work with packages that do not include a composer.json

It has three operational modes allowing for:
1. Default: Searches for a composer.json file in the source code root folder and if found exports the package.
2. Project: Parse project attributes to create a virtual composer.json to use for exporting the package.
3. Advanced: Parse a custom json string supplied in the service configuration to export the package.

It also has branch / tag filters that allows fine-grained filtering of specific branches/tags to be exposed as composer packages. 

In the process of providing a more complex configuration screen i added support for rendering fieldsets, and options in project service configuration along with enabling hints. Also notifications are enabled for Tag/Branch deletion from the gitlab host.

This is and updated merge request for https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/106 working with the latest source code.

Please see feature thread http://feedback.gitlab.com/forums/176466-general/suggestions/5953067-provide-composer-package-repository-support

When a project that has activated the composer service receives a push, the public /packages.json file and project specific json file will be refreshed to include version definitions for the branches and tags as specified in the service configuration in the project's repository.

PR #9301, opened 2015-05-15
Add Composer Project Service (updated #8768)

ikappas:
composer-service-1
# Composer Project Service

This commit adds the per-project option to expose a project as a composer package. It generates and maintains a packages.json file along with a project specific /p/project-{id}.json file (conforming to https://getcomposer.org/doc/05-repositories.md#includes) in the public/ folder, allowing the GitLab install to be used as a Composer repository.

It extends on the original idea providing extended service configuration options that allow to work with packages that do not include a composer.json file.
## Operational Modes

It has three operational modes allowing for:
1. Default: Searches for a composer.json file in the source code root folder and if found exports the package.
2. Project: Parse project attributes to create a virtual composer.json to use for exporting the package.
3. Advanced: Parse a custom json string supplied in the service configuration to export the package.
## Branch / Tag Filters

It also has branch / tag filters that allows fine-grained filtering of specific branches/tags to be exposed as composer packages. 
## References

This is and updated merge request for:
1. https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/106
2. https://github.com/gitlabhq/gitlabhq/pull/8768 

Please see feature thread http://feedback.gitlab.com/forums/176466-general/suggestions/5953067-provide-composer-package-repository-support

When a project that has activated the composer service receives a branch/ tag push, the public /packages.json file and project specific json file will be refreshed to include version definitions for the branches and tags as specified in the service configuration in the project's repository.
## Notes

There is a known issue integrating with the GitLab-Omnibus, namely that the public folder
is owned by 'root' and GitLab is run as 'git'. The GitLab-Omnibus chef script needs to be updated 
to allow access to both packages.json and the /p/ folder in the public/ folder


app/models/project_services/composer_service.rb
```diff
@@ -0,0 +1,471 @@
...
+      { type: 'fieldset', legend: 'Branches:', fields:
+        [
+          { type: 'checkbox',
+            name: 'export_branches',
+            label: 'Branch Listing'
+          },
+          { type: 'text',
+            name: 'branch_filters',
+            label: 'Branch Filters',
+            placeholder: 'branches you wish to export comma seperated.',
```
kojiromike: spelling nitpick: _separated_

ikappas: Thank you for spotting this!

app/models/project_services/composer_service.rb
```diff
@@ -0,0 +1,464 @@
...
+          elsif cjson['name'] != (name_re.match(cjson['name']) || [])[0]
+            error = 'name key must be formatted as "namespace/project"'
+          elsif !cjson['description']
+            error = 'must have a description key specified'
+          elsif cjson['description'].empty?
+            error = 'description key must not be empty'
+          end
+        end
+
+        if error
```
jvanbaarsen: Can you rewrite this as follow?:

```ruby
record.errors.add(attr, error) if error
```

app/models/project_services/composer_service.rb
```diff
@@ -0,0 +1,464 @@
...
+              ['Advanced: use custom JSON', 'advanced']
+            ],
+            default_choice: 'default'
+          },
+          { type: 'select',
+            name: 'package_type',
+            hint: 'Applicable only on project package mode.',
+            choices:
+            [
+              ['AGL Module', 'agl-module'],
```
jvanbaarsen: Where is this list coming from? Is this a Composer specific thing?

app/models/project_services/composer_service.rb
```diff
@@ -0,0 +1,464 @@
...
+      # process packages for all branches
+      project.repository.branches.each do |branch|
+        process_commit(branch)
+      end
+
+      repository.write
+    end
+    update_root_file
+  rescue
+    # skip on error
```
jvanbaarsen: What about log? What kind of errors can occur here? Maybe we want to log those somewhere? At this point we are catching all errors, I think it would be good to narrow it down a bit :-)

app/models/project_services/composer_service.rb
```diff
@@ -0,0 +1,464 @@
...
+  end
+
+  def process_commit(ref)
+    if activated? && ref_exported?(ref)
+      package = ref_package(ref)
+      repository.add_package(package)
+    end
+  rescue
```
jvanbaarsen: Same as the above rescue

jvanbaarsen:
Would it be possible to add some tests to this?

app/models/project_services/composer_service.rb
ikappas: sure

app/models/project_services/composer_service.rb
ikappas: According to composer schema, a package can have type of library, project, metapackage, composer-plugin out of the box while allowing fully customized package types (https://getcomposer.org/doc/04-schema.md#type).

This list comes from the composer/installers package found at https://github.com/composer/installers which is the Multi-Framework Composer Library Installer that supports a wide array of predefined package type installers.

As we allow for full customization of the composer.json via the advanced packaging option I think that having this list for project attributes generated packages is a helper.

jvanbaarsen: Ok, but what about keeping this list up-to-date? How often does it change?

app/models/project_services/composer_service.rb
ikappas: There are two kind of errors that can occur here, namely file system errors and invalid package errors.

The first kind are file system errors that can happen due to file/dir permissions, which is something that needs to be taken care of during the installation of gitlab. Then there could be faulty storage, etc. These type of errors are critical and I would like your suggestion on how to handle them.

The second kind of errors is when we have invalid packages which can happen due to: 
1. unnamed packages, 
2. unsupported version scheme (non schematic)
3. missing/invalid root composer.json when packaging mode is set to default.

These kind of errors are beeing taken care of in the process_commit function and are non critical. Perhaps we should log these in the Application log?.

Please note that I am not an expert ruby developer so your assistance is greatly appreciated.

app/models/project_services/composer_service.rb
ikappas: see above comment.

ikappas:
The underlined functionality of this service relies on the php-composer ruby gem which is a partial port of the composer classes written in php and is tested thoroughly. I have a few tests failing on JsonFile due to inconsistent json handling between php and ruby that I would greatly appreciate your expertise in fixing but all others pass. The source code can be found here: https://github.com/ikappas/php-composer

app/models/project_services/composer_service.rb
ikappas: Well I cant really say how often this list can be changed. That project would probably extend support to additional package types over time but as I mentioned, this is something that can also be achieved using Packagin Mode: Advanced and using a custom.json with any package type the gitlab user wants to use.

Alternatively, we can have this field set as a text input, and users will have to input the package type manually thus not relying to that project at all.

ikappas:
I added the following tests per your request:
/spec/models/project_services/composer_service_spec
/spec/lib/composer/repository/project_repository_spec
/spec/lib/composer/package/loader/project_loader_spec

ikappas:
Is there any feedback/update for this merge request?

mavimo:
:+1: to have it in next release

ikappas:
Hello there... It's been a while and I haven't heard anything from you guys! Any update on this?
Also, I have trouble with the build task failing at bundle exec rake test_ci. Can you please have a look why I am having this issue and elaborate?

Thank you in advance

axilleas:
pinging the pirate crew @jvanbaarsen @DouweM @tsigo @dblessing 

and @maxlazio for the omnibus Note.

axilleas:
So `public/p/` dir should be also part of the backup task, no?

ikappas:
The backup task should include the composer.json file and the /public/p/* folders

stanhu:
FYI, you may need to rebase master to make Semaphore work again.

Lumbendil:
What is pending from this PR to get it merged? I see that Semaphore is passing now.

X-Coder:
Just tested it on the new omnibus update by manually adding your patches.
Works great for me, thank you for making this, it is very useful for me. I hope it will get merged.

app/models/project_services/composer_service.rb
```diff
@@ -0,0 +1,474 @@
...
+        if value.empty?
+          error = 'must be specified'
+        elsif (cjson = ActiveSupport::JSON.decode(value))
+          if cjson.empty?
+            error = 'must not be empty'
+          elsif !cjson['name']
+            error = 'must have a name key specified'
+          elsif cjson['name'].empty?
+            error = 'name key must not be empty'
+          elsif cjson['name'] != (name_re.match(cjson['name']) || [])[0]
```
DouweM: If you add anchors to `name_regex`, like `/\A([A-Za-z0-9&_-]+\/[A-Za-z0-9&_-]+)\z/`, you can do `cjson['name'] =~ name_regex` instead of this weird equality check :)

app/models/project_services/composer_service.rb
```diff
@@ -0,0 +1,474 @@
...
+
+  def description
+    'List your project as a composer package'
+  end
+
+  def help
+    out = 'This project will be publicly listed as a composer package,
+but usage of private and internal repositories will still
+require authentication. '
+    if package_mode == 'default'
```
DouweM: Please make a `case/when` statement out of this.

app/models/project_services/composer_service.rb
```diff
@@ -0,0 +1,474 @@
...
+#
+
+require 'digest'
+require 'composer'
+
+class ComposerService < Service
+  prop_accessor :package_mode, :package_type, :export_branches, :branch_filters,
+                :export_tags, :tag_filters, :custom_json
+
+  validates :package_mode, :package_type, presence: true, if: :activated?
```
DouweM: Please validate that package_mode is one of the three allowed values.

lib/composer/package/loader/project_attributes_loader.rb
```diff
@@ -0,0 +1,56 @@
...
+        end
+
+        # Load a json string or file
+        # Param:  Project project The target gitlab project.
+        # Param:  Branch|Tag ref The target gitlab project branch/tag.
+        # Param:  Hash config A hash containing addional package configuration.
+        # Returns: Composer::Package::Package
+        def load(project, ref, type = 'library')
+          config = {
+            'name'                => project.path_with_namespace.gsub(/\s/, '').downcase,
```
DouweM: Why the `gsub` and `downcase`?

lib/composer/package/loader/project_attributes_loader.rb
```diff
@@ -0,0 +1,56 @@
...
+            'description'         => project.description.nil? ? '' : project.description,
```
DouweM: `project.description || ''` works as well.

lib/composer/package/loader/project_loader.rb
```diff
@@ -0,0 +1,52 @@
...
+        # Returns: Composer::Package::Package
+        def load(project, ref, config)
+
+          config['version'] = parse_version(ref)
+
+          # export distribution package
+          # this works only on public repositories since composer can't handle gitlab oAuth yet.
+          if project.public?
+            config['dist'] = {
+              'url'  => [project.web_url, 'repository', 'archive.zip?ref=' + ref.name].join('/'),
```
DouweM: Please use routing helpers to do this, instead of constructing the path manually.

lib/composer/package/loader/project_loader.rb
```diff
@@ -0,0 +1,52 @@
...
+              'type' => 'zip',
+              'reference' => ref.target
+            }
+          end
+
+          # export source package
+          config['source'] = {
+            'url' => project.url_to_repo,
```
DouweM: How will this work with private repos that need authentication?

lib/composer/package/loader/project_loader.rb
```diff
@@ -0,0 +1,52 @@
...
+            'reference' => ref.target
+          }
+
+          @loader.load(config)
+        end
+
+        private
+
+        def parse_version(ref)
+          ref.instance_of?(Gitlab::Git::Branch) ? "dev-#{ref.name}" : ref.name
```
DouweM: What's the idea behind this?

lib/composer/package/loader/project_attributes_loader.rb
```diff
@@ -0,0 +1,56 @@
+module Composer
```
DouweM: I think it's dangerous to move into Composers namespace territory like this. Please house these under `Gitlab::Composer`.

app/models/project_services/composer_service.rb
```diff
@@ -0,0 +1,474 @@
...
+  def title
+    'Composer'
+  end
+
+  def description
+    'List your project as a composer package'
+  end
+
+  def help
+    out = 'This project will be publicly listed as a composer package,
```
DouweM: I think it needs to be made much clearer that enabling this will make information about this private repo public. I can also see GitLab admins wanting to disable this service to prevent that from happening.

app/models/project_services/composer_service.rb
```diff
@@ -0,0 +1,474 @@
...
+  def root_filename
+    'packages.json'
+  end
+
+  def root_path
+    File.join(output_dir, root_filename)
+  end
+
+  def repo_filename
+    "project-#{project.id}.json"
```
DouweM: Are these files cleaned up when the project is destroyed? Are these files updated when the project's name, path, description, whatever changes?

app/models/project_services/composer_service.rb
```diff
@@ -0,0 +1,474 @@
...
+
+  def supported_events
+    %w(push tag_push)
+  end
+
+  def execute(push_data)
+    process_project
+  end
+
+  def process_project
```
DouweM: Please move this to a separate class in lib. The service is way too heavy this way.

app/models/project_services/composer_service.rb
```diff
@@ -0,0 +1,474 @@
...
+  end
+
+  def repository
+    @repository ||= Composer::Repository::ProjectRepository.new(
+      Composer::Json::JsonFile.new(repo_path)
+    )
+  end
+
+  def ref_exported?(ref)
+    if ref.instance_of?(Gitlab::Git::Branch)
```
DouweM: Please use a case/when statement for this?

app/models/project_services/composer_service.rb
```diff
@@ -0,0 +1,474 @@
...
+  end
+
+  def branch_exported?(branch)
+    if branch_filters
+      filters = branch_filters.strip! || branch_filters
+      filters = filters.gsub(' ', '').split(',')
+    else
+      filters = []
+    end
+    if filters.empty?
```
DouweM: How about replacing the `if branch_filters` stuff by `filters = branch_filters.split(',').map(&:strip)`.

app/models/project_services/composer_service.rb
DouweM: Would it be reasonable to disallow using this service at all for private projects?

app/models/project_services/composer_service.rb
DouweM: Are these file backed up? Should they be?

DouweM:
@ikappas Thanks for your work on this! I have a couple of concerns, but I like the general idea.

lib/composer/package/loader/project_loader.rb
Lumbendil: Afaik it's meant to go this way, so the repos still require access and no-one can access the code.

app/models/project_services/composer_service.rb
Lumbendil: Talking from my own case. I'd prefer if this file was behind Basic Auth or similar. It would not be reasonable to disallow this service for all private projects since it's quite likely that's what people would want to use it for, at least I want it for generating a packages.json file out of my private repos, which are used by other private repos :)

X-Coder:
Currently you are using a static list of selectable package types. I have some projects which uses custom composer installers that depend on custom package types (see https://getcomposer.org/doc/04-schema.md#type) not in the list.
Is there a way you could add a input text field for specifying these custom package type instead of one in the list?
Maybe its a good idea to allow admins to extend the list too?

lib/composer/package/loader/project_loader.rb
Lumbendil: Tags are given to composer as they are, but branch versions always get the "dev-" appended. It's simplified (branches which look like versions also should use "ref.name") but it should do the trick most of the time. https://getcomposer.org/doc/articles/aliases.md#why-aliases-

DouweM: All right, thanks.

app/models/project_services/composer_service.rb
DouweM: Does Composer support repositories with Basic Auth?

X-Coder: Yes, it supports also ssh, ssl with client side certificate, or a custom HTTP Header field for token authentication. See https://getcomposer.org/doc/articles/handling-private-packages-with-satis.md#security.

In my case I use private repositories together with ssh using public key auth. So the generated packages.json has to list private repositories in this case. But someone else might find it useful not to list private repositories at all, maybe meaning just disable the composer service for this project.

app/models/project_services/composer_service.rb
ikappas: This is fixed and pending commit.

app/models/project_services/composer_service.rb
ikappas: This is fixed and pending commit.

app/models/project_services/composer_service.rb
ikappas: This is fixed and pending commit.

lib/composer/package/loader/project_attributes_loader.rb
ikappas: if project.path_with_namespace is guaranteed to be space free and downcased this is not required.

lib/composer/package/loader/project_attributes_loader.rb
ikappas: This is fixed and pending commit.

lib/composer/package/loader/project_loader.rb
ikappas: As @Lumbendil mentioned public projects expose both distribution (archive url) and source information (gitlab repo url) so one can use the archive over the gitlab repo from composer which is faster and more efficient.

When a project is private, what is exposed is meta about this project such as name, description, source url, reference, version, etc. in order to let composer know that we have a package available at a specific version.

Accessing the package is done via composer by performing a git clone so in private repos this requires gitlab authentication (ssh, ssl, etc). 
see @X-Coder note below

lib/composer/package/loader/project_loader.rb
ikappas: Some assistance here would be much appreciated!

lib/composer/package/loader/project_loader.rb
ikappas: I am covered by @Lumbendil answer.

app/models/project_services/composer_service.rb
ikappas: I would also like to mention that work is being done https://github.com/composer/composer/pull/3765 to introduce a Gitlab Driver in composer.

app/models/project_services/composer_service.rb
ikappas: Right before a project is destroyed the /p/project-id.json should be deleted and projects.json updated to remove the package entry. I would like your input how to achieve this within composer-service.

Also which events can I use in order to change the project name, path or description whenever it changes?

As far as backing these files is concerned, these should always be in sync with the service configuration so they should be backed up.

app/models/project_services/composer_service.rb
ikappas: This is fixed and pending commit.

app/models/project_services/composer_service.rb
ikappas: This is fixed and pending commit.

lib/composer/package/loader/project_attributes_loader.rb
ikappas: I tried moving the classes to Gitlab::Composer namespace but I can't get them to load their parent classes. Alternatively I can prefix the classes with Gitlab_______ . Let me know if this works for you.

ikappas:
@X-Coder 
The static list of selectable package types is from https://github.com/composer/installers 

There are two issues here. One is that this list needs to be manually updated in the source code whenever it changes. The other is that it does not allow for direct customization.

The only way to use custom package type is  to use the custom json packaging mode and use a fully customized json.

I could add more configuration options, as per your example, allowing package type customisation but the service form rendering does not support more advanced ui configuration such as conditional visibility dependant on input selection. Thus I feel that it may clutter the whole composer service settings.

ikappas:
@DouweM 
Thank you for your feedback. I have a couple of questions/clarifications and require some assistance to fix/update the rest of your suggestions as mentioned on each issue. I would really like to have this ready for the Gitlab 8.0 milestone so your help is greatly appreciated!

lib/composer/package/loader/project_attributes_loader.rb
DouweM: @ikappas It's guaranteed to be space-free, it's not guaranteed to be downcased. Is being downcased a requirement for Composer?

lib/composer/package/loader/project_loader.rb
DouweM:
```ruby
Rails.application.routes.url_helpers.archive_namespace_project_repository_path(project.namespace, project, ref: ref, format: 'zip')
```

lib/composer/package/loader/project_attributes_loader.rb
DouweM: If you do `Gitlab::Composer`, you need to reference the original `Composer` as `::Composer`, where `::` is the "root namespace". Just `Composer` will refer to `Composer` in the current namespace, which is `Gitlab` since you're inside `module Gitlab`.

app/models/project_services/composer_service.rb
DouweM: @ikappas I think you missed this one :)

lib/composer/package/loader/project_attributes_loader.rb
X-Coder: Don't know if it is specified somewhere, but each composer package on packagist is also lowercased and space free. See https://packagist.org/explore/popular. I would say it is a good idea to make it packagist compatible right from the start.

app/models/project_services/composer_service.rb
DouweM: @ikappas If that goes through, is this MR needed at all? With the lack of authentication on private repo information, the package files not automatically updating in response to project changes, and package files not being backed up, I'm a bit wary about the current implementation. All of these are next to impossible to do using services right now.

lib/composer/package/loader/project_attributes_loader.rb
DouweM: @X-Coder My main concern is that in that case the composer name will no longer match the path in the GitLab URL. So `gitlab.com/SomeProject/CoolLibrary` will become `someproject/coollibrary`, and they can't just copy it from the GitLab URL. But if lowercase is the norm, let's go with that.

X-Coder: I see only one issue in it, what happens if someone creates gitlab.com/SomeProject/CoolLibrary and someone else gitlab.com/someproject/coolLibrary. This leads in conflict.

DouweM: @X-Coder Usernames and groups uniqueness is case insensitive, but we do preserve case, so that conflict cannot exist.

ikappas: @DouweM Since the names are guaranteed to be space-less then I will remove the unnecessary gsub and keep the downcase. 

Also let me point out that the package name doesn't need to match the path of the Gitlab repo and can be arbitrarily set using the custom json option to any namespace/package_name.

The "source" and "dist" keys are used to fetch the package from the correct project url.

lib/composer/package/loader/project_loader.rb
ikappas: Thank you. Will have a go at it tommorow.

app/models/project_services/composer_service.rb
DouweM: @ikappas Could you respond to my comment here?

lib/composer/package/loader/project_attributes_loader.rb
ikappas: Thank you. I was missing the :: root namespace and was indeed running into a namespace conflict. Will have a go at it tommorow.

app/models/project_services/composer_service.rb
ikappas: @DouweM I think the work being done there is to allow composer to access a gitlab repo directly using Oauth and has also been pending for a while. (I hope it gets merged eventually)

This MR is enabling Gitlab to act as a packagist repository. I admit that there are many ways to tackle this. For example one can setup satis and use that configuration to serve up packages.

For smaller shops, like mine, doing so would be an overkill since we have a small amount of packages and having this kind of gitlab service eliminates the need to use a dedicated solution while only having to maintain our gitlab server.

I agree that this solution has room for improvement and would be more than happy to improve on it. Gitlab services have been engineered mainly under a "remote" point of view and lack in "local" perspective and are perhaps in need of some redesign. 

The renaming/changing project attributes issue affects not only this service but even Gitlab CI. What happens if you rename/move a project that is linked to Ci. From my experience I can tell you that the Ui links pointing back from CI -> Gitlab don't get updated and 404. I had to manually query the ci database and change the project paths in order to fix them. So events for these kinds of changes should be implemented so that services can sync and propagate the changes.

Let me know what you think.

app/models/project_services/composer_service.rb
ikappas: @DouweM I kind of did. :) Wanted to have a look at the others first.

lib/composer/package/loader/project_attributes_loader.rb
ikappas: This is fixed and pending commit.

lib/composer/package/loader/project_loader.rb
ikappas: This is fixed and pending commit.

lib/composer/package/loader/project_attributes_loader.rb
ikappas: This is fixed and pending commit.

ikappas:
@X-Coder I implemented custom package type per your suggestion. Hope it helps!

X-Coder:
@ikappas Thank you very much for implementing the custom package type, I just tested it and it works great.

ikappas:
@X-Coder No problem! That is good to hear.

@DouweM I have just implemented the backup rake task needed for backing/restoring the service files. In the omnibus installation there is a problem with permissions that needs to be addressed in the similar way as the uploads folder. 

The solution for the omnibus package team is as follows:

1. create a composer folder for housing the files under the /var/opt/gitlab/gitlab-rails folder.
2. move the packages.json file from /opt/gitlab/embedded/service/gitlab-rails/public/ to  /var/opt/gitlab/gitlab-rails/composer folder
3. move the "p" folder from /opt/gitlab/embedded/service/gitlab-rails/public/ to the  /var/opt/gitlab/gitlab-rails/composer folder.
4. symlink /var/opt/gitlab/gitlab-rails/composer/packages.json file to /opt/gitlab/embedded/service/gitlab-rails/public/packages.json
5. symlink /var/opt/gitlab/gitlab-rails/composer/p folder to /opt/gitlab/embedded/service/gitlab-rails/public/p

This enables the git user to be able to write both to /public/packages.json and the /public/p folder.

PS: The build fails for some reason... I will have a look at it on the weekend.

app/models/project_services/composer_service.rb
ikappas: Backup rake task is committed. The events for updating the project attributes are pending.

app/models/project_services/composer_service.rb
DouweM: @ikappas Thanks for you response, I'll get back to you early next week.

peuh:
@ikappas Having this feature in Gitlab CE would be awesome as it would save us a lot of time seting up Satis and integrating it with our CI tool. 
This PR hasn't seen any activity in almost a couple of months so I was wondering if you were still planning to have it merged in CE. Any update?

MattLabatte:
:+1: Would really appreciate if it could be merged soon

AlexDubstone:
Agree, having this directly in Gitlab would be a huge win. Any chance to see this coming soon?

Baldinof:
:+1: It would be awesome for PHP developpers

ikappas:
@peuth, @MattLabatte, @AlexDubstone, @Baldinof I just updated the PR to the latest master and fixes some API related errors. Until this PR gets merged, if ever(it's now 6 months pending + about 3months the previous PR, + even more the one before that!!!!) you can manually patch your gitlab installation to get the functionallity.

AlexDubstone:
@ikappas Awesome! Thank you a lot for your work and endurance. Really hope to see this getting merged soon. The attention your PR got lately definitely shows that a lot of people would profit from this feature. 
Thanks, once again!

ikappas:
@AlexDubstone No problem! I really hope so as I too find this functionality extremely helpful as a PHP developer.

dada87:
Yes can't wait eather for it to be merged ! Thnaks for this great work.

peuh:
@ikappas Thanks for the update, considering that all the tests pass, hopefully,  the gitlab team will finally accept it :-) Thanks for all the time you've put in this feature. Really appreciate!

DouweM:
My apologies for the delay, and thanks for bringing this on my radar again. I will review it again asap.

ikappas:
Can somebody help out in pointing out why this build failed twice on code that Is not relevant to this PR and succeeds building on master ?!?

master build:
https://semaphoreci.com/gitlabhq/gitlabhq/branches/master/builds/1929

last two builds:
https://semaphoreci.com/gitlabhq/gitlabhq/branches/pull-request-9301/builds/38
and 
https://semaphoreci.com/gitlabhq/gitlabhq/branches/pull-request-9301/builds/39

ikappas:
The third fast forward passed the build... semaphore must not be working properly all the time...

Lumbendil:
@ikappas would it be possible to add some sort of http basic auth to this? I know the code itself should be safe because it's checked out from the repository, but it'd be good to be able to add it anyway :)

ikappas:
@Lumbendil Thank you for your input. I think this PR has enough changes as it is, for the Gitlab guys to review and merge. 

As you also mentioned, the composer service works effectively as it is, since it allows composer to download the code archive on public projects and requires authenticating to the gitlab instance for pulling private/protected code.

I hope this gets merged soon so that we can all benefit as php developers from this service and we can later consider adding new features such as basic auth -- which can probably be done manually atm.

doc/api/services.md
```diff
@@ -173,6 +173,45 @@ Get Campfire service settings for a project.
 GET /projects/:id/services/campfire
 ```
 
+## Composer
+
+Simple web-based real-time group chat
```
mavimo: Maybe this description should be changed ;)

AlexDubstone:
Did this land off the radar again? Any chance to see this PR ever get merged? It's a really useful feature and would enhance Gitlab a lot for PHP devs.

peuh:
@DouweM Any update on this?

DouweM:
My apologies for the delay guys, my "asap" isn't what it used to be. As you may have seen, we just released [GitLab 8.2](https://about.gitlab.com/2015/11/22/gitlab-8-2-released/), which is a very big release that kept all of us very busy this last month.

This issue still on my radar, and I really hope to get to it this week.

lib/gitlab/composer/repository/project_repository.rb
```diff
@@ -0,0 +1,84 @@
...
+            crc = Digest::CRC32.hexdigest(packages[0].name)
+            packages.each do |package|
+              next if package.instance_of?(::Composer::Package::AliasPackage)
+              data[package.pretty_name] = {} unless data[package.pretty_name]
+              data[package.pretty_name][package.pretty_version] = @dumper.dump(package)
+              data[package.pretty_name][package.pretty_version]['uid'] = "#{crc}" + (uid += 1).to_s
+            end
+          end
+          @file.write({ 'packages' => data })
+        rescue Exception => e
```
DouweM: Please don't rescue from `Exception`: http://stackoverflow.com/questions/10048173/why-is-it-bad-style-to-rescue-exception-e-in-ruby

DouweM: This method can be simplified like this:

```ruby
def write
  return if packages.nil? || packages.empty?

  data = {}
  
  uid = 0
  crc = Digest::CRC32.hexdigest(packages[0].name)
  packages.each do |package|
    next if package.instance_of?(::Composer::Package::AliasPackage)
    uid += 1

    version_data = @dumper.dump(package)
    version_data['uid'] = "#{crc}-#{uid}"
    
    package_data = data[package.pretty_name] ||= {}
    package_data[package.pretty_version] = version_data
  end
  
  @file.write({ 'packages' => data })
rescue => e
  Gitlab::AppLogger.error("ProjectRepository: #{e.message}")
end
```

app/models/project_services/composer_service.rb
```diff
@@ -0,0 +1,516 @@
...
+    update_root_file
+
+  end
+
+  def process_commit(ref)
+    if activated? && ref_exported?(ref)
+      package = ref_package(ref)
+      repository.add_package(package)
+    end
+  rescue Exception => e
```
DouweM: Please don't rescue from `Exception`: http://stackoverflow.com/questions/10048173/why-is-it-bad-style-to-rescue-exception-e-in-ruby

app/models/project_services/composer_service.rb
```diff
@@ -0,0 +1,516 @@
...
+    if activated?
+      includes[key] ||= {}
+      includes[key]['sha1'] = Digest::SHA1.file(repo_path).hexdigest
+      includes = includes.sort_by{|k,v| k}.to_h
+    else
+      includes.delete(key)
+    end
+
+    root_json.write({ packages: [], includes: includes })
+  rescue Exception => e
```
DouweM: Please don't rescue from `Exception`: http://stackoverflow.com/questions/10048173/why-is-it-bad-style-to-rescue-exception-e-in-ruby

lib/gitlab/composer/repository/project_repository.rb
```diff
@@ -0,0 +1,84 @@
...
+            unless data.is_a?(Hash)
+              raise ::Composer::UnexpectedValueError,
+                    'Could not parse package list from the repository'
+            end
+            unless data['packages'].is_a?(Hash)
+              raise ::Composer::UnexpectedValueError,
+                    'Could not parse package list from the repository'
+            end
+            packages = data['packages']
+          rescue Exception => e
```
DouweM: Please don't rescue from `Exception`: http://stackoverflow.com/questions/10048173/why-is-it-bad-style-to-rescue-exception-e-in-ruby

lib/api/entities.rb
```diff
@@ -251,9 +251,18 @@ class ProjectService < Grape::Entity
...
       # Expose serialized properties
       expose :properties do |service, options|
-        field_names = service.fields.
-          select { |field| options[:include_passwords] || field[:type] != 'password' }.
-          map { |field| field[:name] }
+        field_names = []
+        service.fields.each do |field|
+          if field[:type] == 'fieldset'
+            field[:fields].each do |subfield|
+              if options[:include_passwords] || subfield[:type] != 'password'
```
DouweM: Can we get rid of this duplication using a helper method we call recursively?

lib/backup/composer.rb
```diff
@@ -0,0 +1,66 @@
+require 'open3'
+
+module Backup
+  class Composer
```
DouweM: Can we reuse `Backup::Files` here, like `Backup::Uploads` does?

lib/gitlab/composer/package/loader/project_attributes_loader.rb
```diff
@@ -0,0 +1,58 @@
...
+            commit = project.repository.commit(ref.target)
+            commit.committed_date.strftime('%Y-%m-%d %H:%M:%S')
+          rescue
+            # If there's a problem, just skip the "time" field
+          end
+
+          def parse_keywords(project)
+            project.tags.collect { |t| t['name'] }
+          rescue
+            # If there's a problem, just skip the "keyworks"
```
DouweM: What problem would there be?

lib/gitlab/composer/package/loader/project_attributes_loader.rb
```diff
@@ -0,0 +1,58 @@
...
+            super(project, ref, config)
+          end
+
+          private
+
+          def parse_time(project, ref)
+            commit = project.repository.commit(ref.target)
+            commit.committed_date.strftime('%Y-%m-%d %H:%M:%S')
+          rescue
+            # If there's a problem, just skip the "time" field
```
DouweM: Can we skip this and simply check if `commit` isn't `nil`?

lib/gitlab/composer/package/loader/project_loader.rb
```diff
@@ -0,0 +1,54 @@
...
+          # Param:  Project project The target gitlab project.
+          # Param:  Branch|Tag ref The target gitlab project branch/tag.
+          # Param:  Hash config A hash containing addional package configuration.
+          # Returns: Composer::Package::Package
+          def load(project, ref, config)
+
+            config['version'] = parse_version(ref)
+
+            # export distribution package
+            # this works only on public repositories since composer can't handle gitlab oAuth yet.
```
DouweM: I think it can now!

lib/gitlab/composer/repository/project_repository.rb
```diff
@@ -0,0 +1,84 @@
...
+module Gitlab
+  module Composer
+    module Repository
+      class ProjectRepository < ::Composer::Repository::WritableHashRepository
+
+        # Initializes the project repository.
+        # @param [Composer::Json::JsonFile] repository_file The repository json file
+        def initialize(repository_file)
+          unless repository_file
+            raise ::Composer::ArgumentError,
```
DouweM: Why do we use errors from `::Composer` here rather than just using the Ruby ones?

lib/gitlab/composer/repository/project_repository.rb
```diff
@@ -0,0 +1,84 @@
...
+                  'repository_file must be specified'
+          end
+          unless repository_file.is_a?(::Composer::Json::JsonFile)
+            raise ::Composer::TypeError,
+                  'repository_file type must be a \
+                  Composer::Json::JsonFile or superclass'
```
DouweM: *subclass

DouweM:
@ikappas Thanks for your patience. I reviewed the code; mostly issues with the Ruby style, nothing major about the structure, so good job! :)

Could you please fix those issues and the merge conflicts? It would also be nice if you could (cross-)post this to https://gitlab.com/gitlab-org/gitlab-ce/merge_requests and see if GitLab CI has any additional complaints; the Semaphore build script is a bit behind.

lib/gitlab/composer/repository/project_repository.rb
ikappas: @DouweM :  I fixed the rescue Exception => e methods and pending commit.

The write method should write the {"packages":{}} json string to the repository file. Your suggestion skips the writing altogether. :)

app/models/project_services/composer_service.rb
ikappas: @DouweM  Fixed. Pending commit

app/models/project_services/composer_service.rb
ikappas: @DouweM  Fixed. Pending commit

lib/gitlab/composer/repository/project_repository.rb
ikappas: @DouweM  Fixed. Pending commit

lib/gitlab/composer/repository/project_repository.rb
DouweM: @ikappas How do you mean? The `@file.write({ 'packages' => data })` is still in there.

lib/gitlab/composer/repository/project_repository.rb
```diff
@@ -0,0 +1,84 @@
...
+        def write
+          data = {}
+          unless packages.nil? || packages.empty?
+            uid = 0
+            crc = Digest::CRC32.hexdigest(packages[0].name)
+            packages.each do |package|
+              next if package.instance_of?(::Composer::Package::AliasPackage)
+              data[package.pretty_name] = {} unless data[package.pretty_name]
+              data[package.pretty_name][package.pretty_version] = @dumper.dump(package)
+              data[package.pretty_name][package.pretty_version]['uid'] = "#{crc}" + (uid += 1).to_s
```
DouweM: What exactly is the point of this UID, and why does it only use the name for the first package, and then a random counter?

lib/gitlab/composer/repository/project_repository.rb
mavimo: @DouweM i suppose on `return if packages.nil? || packages.empty?` that skip the `@file.write({ 'packages' => data })` code.

DouweM: @mavimo Ah you're right, I missed that.

lib/backup/composer.rb
ikappas: @DouweM I had a look at that while developing Backup:Composer but Backup:Files only takes a single folder as target. We need to pickup both the /public/packages.json file and the /public/p/ folder into the backup tar. This would require that we refactor the Backup:Files class to accept an array of targets and complicate the class significantly.

lib/gitlab/composer/package/loader/project_attributes_loader.rb
ikappas: @DouweM If the ref passed to this file is invalid, then this function would fail and since the time meta is optional this can be safely return nil an subsequently be skipped.

lib/gitlab/composer/package/loader/project_attributes_loader.rb
ikappas: @DouweM I could rewrite this as:
```
return commit.committed_date.strftime('%Y-%m-%d %H:%M:%S') unless commit.nil?
```

lib/gitlab/composer/package/loader/project_loader.rb
ikappas: @DouweM  Yes it was merged last week!!!

lib/gitlab/composer/repository/project_repository.rb
ikappas: since we are subclassing ::Composer::Repository::WritableHashRepository I used the ::Composer errors for compatibility.

lib/gitlab/composer/repository/project_repository.rb
ikappas: @DouweM Fixed and pending commit.

doc/api/services.md
ikappas: @mavimo Thanks for spotting that! Fixed and pending commit.

lib/api/entities.rb
ikappas: I would like your assistance with this ... where would I put the helper function and how would I call it here?
```
def get_field_names(field, options)
  field_names = [] 
  if field[:type] == 'fieldset'
    field[:fields].each do |sub_field|
      sub_field_names = get_field_names(sub_field, options);
      field_names << sub_field_names unless sub_field_names.empty?
      field_names.flatten!
    end
  else 
    field_names << field[:name] if options[:include_passwords] || field[:type] != 'password'
  end
  field_names
end
```

DouweM: @ikappas Can you do something similar to `def service_attributes` in `lib/api/helpers.rb`?

lib/backup/composer.rb
DouweM: @ikappas Would it be possible to subclass `Backup::Files`, and override all its methods to do the default think for `public/p/`, and _also_ backup `public/packages.json`? I agree that `Backup::Files` doesn't (currently) need to support an array of targets.

lib/gitlab/composer/package/loader/project_attributes_loader.rb
DouweM: I think that's better, with `if commit` instead of `unless commit.nil?`, and without the explicit `return` :)

lib/gitlab/composer/package/loader/project_attributes_loader.rb
DouweM: I was talking about the `parse_keywords` method, nothing can really go wrong there, can it?

lib/gitlab/composer/repository/project_repository.rb
DouweM: OK, makes sense.

lib/gitlab/composer/package/loader/project_attributes_loader.rb
ikappas: I guess project.tags should be an empty array if not tags exists so collect would not thow...
Fixed and pending commit!

lib/gitlab/composer/package/loader/project_attributes_loader.rb
ikappas: Fixed and pending commit!

lib/gitlab/composer/repository/project_repository.rb
ikappas: @DouweM This UID was required by composer to identify the packages in the repository. I can't find relevant documentation on the uid so I guess its obsolete and removed it.
Fixed and pending commit!

lib/backup/composer.rb
ikappas: Sure I can do that.

lib/api/entities.rb
ikappas: After going about it for half a day... 
The entities.rb contain GrapeEntity classes for which I can't find how to access the lib/api/helpers.rb defined helper method. If that is possible can you please let me know how to do so and I can implement it.

ikappas:
@DouweM 

> Could you please fix those issues and the merge conflicts? It would also be nice if you could (cross-)post this to https://gitlab.com/gitlab-org/gitlab-ce/merge_requests and see if GitLab CI has any additional complaints; the Semaphore build script is a bit behind.

I haven't ever done that so I would be more than happy to cross-post provided you can give me some assistance :)

Also a not so quick question:
I've read that Gitlab-EE 8.2 has mirroring and would like to know if that can ever be included in Gitlab-CE.

It would really fit nicely with this MR since we could mirror php packages and expose them from the composer service eliminating the need for a package proxy altogether!. 

Actually this is what we have set up in our small company for our own developed packages and mirrored packages but with scripts and cron (very hacky). We would be more than happy to get the Gitlab-EE but at the current cost for 10 seats for a small startup with 3 devs is not  really an option.

What do you think?

lib/api/entities.rb
ikappas: In addition, subfields should not iterate recursively. 
I've come to think that this function is way more concise and maintainable *as is* with this duplication within the span of three lines compared to any other alternative involving a helper method, recursion, nesting level etc.

DouweM: @ikappas All right, let's keep it like this.

DouweM:
@ikappas All right, I'll close this PR and we can move the discussion to GitLab.com.

As for repository mirroring, we currently don't have any plans to bring this feature to CE. Since GitLab.com runs EE, it's also an option to set up the repository mirroring there and use GitLab.com as your Composer repo. Would that work for you?

ikappas:
@DouweM Thank you. Feel free to close this PR and will continue on Gitlab.com.

As for repository mirroring, I am hoping you reconsider making plans to bring this feature to CE. As far as moving all our repos to GitLab.com we would have to move our 20+ groups containing about 100+ projects off-site to Gitlab.org and restructure them so ... not really.

ccoVeille:
@ikappas @DouweM By "will continue on Gitlab.com" you mean here https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/1921 right ?

ikappas:
@ccoVeille That's right! +1:

PR #9317, opened 2015-05-21
Moved changelog stuff to the correct release

jvanbaarsen:
**Why was this needed?**
The changes are pulled in 7.12.0 but where mistakenly put under 7.11.0


maxlazio:
:+1:

Issue #5625, opened 2013-11-19
Cannot save whole bio on Edit-Profile-Page when entering 'special' characters


pmuens:
## Summary:
When I add a special character (such as '<') in the bio-Field on the Edit-Profile-Page and click 'save', my bio gets truncated.

## Steps to reproduce:
Open the Edit-Profile-Page. Type in the string 'I <3 programming'. and click 'save'

## Expected behavior:
The string gets stored in the database and will be displayed after the save in the text box for the bio.

## Observed behavior:
The string gets only stored up to the first occurrence of '<'. only 'I ' gets stored. What is more confusing is, that when one replace the character '<' with '>' ('I >3 programming') , '>' gets replaced with '&gt'.

I'm using your GitLab cloud.

Great job BTW!

jvanbaarsen:
Still in 7.10.4-ee d17bba9

PR #9318, opened 2015-05-21
Allow special characters in users bio

jvanbaarsen:
**What does this do?**
It removes the very strict sanitation on the users bio field, so that people can
have a bio like "I <3 GitLab"

**Why is this needed?**
Currently when you enter a bio with "I <3 GitLab", we only store "I ". This is
unexpected behaviour, since we want users to have a normal profile, without
having to worry what characters are allowed and which are not.

**Related issues:**
Fixes https://github.com/gitlabhq/gitlabhq/issues/5625

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -37,6 +37,9 @@ v 7.11.1
 v 7.11.0 
   - Fall back to Plaintext when Syntaxhighlighting doesn't work. Fixes some buggy lexers (Hannes Rosenögger)
   - Get editing comments to work in Chrome 43 again.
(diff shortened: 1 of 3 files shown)


randx:
@jvanbaarsen can people inject XSS via this field if we remove sanitize?

jvanbaarsen:
@randx If you mean stuff like: `<script>alert('hi');</script>` then no. I checked that.

randx:
@jvanbaarsen ok cool. Lets merge this one

Issue #8565, opened 2015-01-12
Wiki: Page Name with period 404s


VertigoRay:
# Components

```text
GitLab 7.6.2
GitLab Shell 2.4.0
GitLab API v3
Ruby 2.1.5p273
Rails 4.1.1
```

# Summary

When working with a page name with a period, the page cannot be found unless you include the `.markdown` extension

# Description

The list at my wiki pages list (ie: `https://gitlab.vertigion.com/project/repo/wikis/pages`) shows my documentation page: *MyApp.exe*. Clicking *MyApp.exe* will take you to: `https://gitlab.vertigion.com/project/repo/wikis/MyApp.exe`. This `/wikis/MyApp.exe` page brings up the creation page, as if the file doesn't exists. If you put anything in the field and click *Create Page*, you will get an error message: `Duplicate page: Cannot write //MyApp.exe.markdown, found //MyApp.exe.markdown.`.  Changing the URL to `MyApp.exe.markdown` displays the existing wiki article.

sque:
I can also confirm the bug.

Nice workaround!

Issue #8166, opened 2014-10-25
404 Error when creating Wiki page with spaces


demisx:
I know the hint says _don't use spaces_ when creating a new Wiki page, but if I forget about it and do use spaces, it gives me a 404 error page after clicking on "New page" button. A better UX approach would be to either allow spaces and then replace them with hyphens upon submission or show a validation error if spaces are present. A user should never get to the 404 page.

demisx:
+1 Still happening in the latest 7.10 version. Need just basic form validation for now. Long term, we should be able to create pages with any name, of course.

demisx:
This is still happening in 7.11. I've used `generate csr` as page slug.

Razer6:
This will be available with 7.12

Issue #5357, opened 2013-10-17
Colon's ":" in wiki page titles cause wiki/pages to throw a RoutingException


karlhungus:
To reproduce:

1. Create project/ edit wiki
2. use gollum check out project, create page title "Foo: Bar"
3. push changes back
4. click pages
```
ActionController::RoutingError at /test1/test/wikis/pages
No route matches {:action=>"show", :controller=>"projects/wikis", :project_id=>#<Project id: 1, name: "test", path: "test", description: "", created_at: "2013-10-16 14:28:10", updated_at: "2013-10-16 14:32:29", creator_id: 2, default_branch: "foo", issues_enabled: true, wall_enabled: f...
```

Issue #3913, opened 2013-05-11
I cannot Enter Chinese title in wiki inside gitlab


jw2007like:
I cannot Enter Chinese title in wiki inside gitlab.

I was a programmer in mainland China. My mother language is Chinese.

Our team is migrating from svn to git. Many colleagues are learning git. I try to use gitlab set up a git server. I found a problem, is wiki system inside gitlab, can not enter the Chinese title. For example, I wanted to write a "How to setup” document. How to setup in Chinese is “如何安装”。

1.I try enter “如何安装”in wiki.
![01](https://f.cloud.github.com/assets/4402641/491348/3a820bbc-ba19-11e2-8a22-54a98bbe8c61.jpg)

2.wiki 404 error
![02](https://f.cloud.github.com/assets/4402641/491349/90b0d0ae-ba19-11e2-9fd2-dd918153a384.jpg)

3.I can write English title in wiki inside gitlab.
![03](https://f.cloud.github.com/assets/4402641/491350/97b0f0c8-ba19-11e2-8130-96be09af58ec.jpg)

4.It works
![04](https://f.cloud.github.com/assets/4402641/491351/9ccddd0a-ba19-11e2-8368-6f2cb3fae34b.jpg)

1.I use wiki in http://bitbucket.org/。enter Chinese title. It works.
![05](https://f.cloud.github.com/assets/4402641/491352/a3bdd552-ba19-11e2-8f4f-d97d0f2758bc.jpg)

2.wiki system in bitbucket.org .Enter Chinese title.It works.
![06](https://f.cloud.github.com/assets/4402641/491353/a6a56b22-ba19-11e2-9c5d-55a14633bbb9.jpg)

question

 I use gitlab version 5.0.How can I enter Chinese title in wiki inside gitlab?

0oneo:
any plan to solve this?

JiffsMaverick:
Same issue for Russian language on 7.9-stable and 7.10-stable.

toplinuxsir:
Same issue for gitlab 8.5.3!
The problem still exist!

lightinen:
8.5.4 the same

stanhu:
This is fixed in GitLab 8.6.

jyunfan2015:
yes,i try to use,it works. My gitlab version is 

System information
System:         Ubuntu 16.04
Current User:   git
Using RVM:      no
Ruby Version:   2.3.3p222
Gem Version:    2.6.6
[...]

PR #9319, opened 2015-05-21
Validate wiki page creation

jvanbaarsen:
**What does this do?**
It adds validation to the creation of a wiki page, that way the user gets real
feedback instead of just a 404 page if the name of the wiki page was invalid

**Why is this needed?**
There are a lot of characters that are not allowed in the creation of a wiki
page, there is even a small text that is saying: Please don't use spaces.
Although we have that text there, we don't actually validate on this. This
commit adds validation on the title and gives the user actual feedback.

**What issues does this fix?**
Fixes http://github.com/gitlabhq/gitlabhq/issues/5357
Fixes https://github.com/gitlabhq/gitlabhq/issues/8565
Fixes https://github.com/gitlabhq/gitlabhq/issues/3913
Fixes https://github.com/gitlabhq/gitlabhq/issues/8166


Razer6:
@jvanbaarsen Can you rebase?

randx:
@jvanbaarsen feel free to merge

PR #9322, opened 2015-05-22
Git gc as rescheduling sidekiq background job.

per-garden:
Running git garbage collection (git gc) at regular intervals for all repositories managed by Gitlab.

git_gc


jvanbaarsen:
@per-garden Did you consider using a crontab instead of scheduling this in Sidekiq?

per-garden:
Yup. Using crontab would require yet another setup not apparently related to Gitlab though. I don't think this is preferable. Since Gitlab's concept is that it "owns" all repositories, I think an administrator in Gitlab should be able to monitor this kind of task. Thus a common point of monitoring and configuration for all git work in one place.

jvanbaarsen:
@per-garden What do you think about https://github.com/javan/whenever? That way you don't have to setup another crontab, we can do it from the codebase, but still we are not using Sidekiq for this.

rumpelsepp:
> but still we are not using Sidekiq for this.

What is the disadvantage of using sidekiq for these kind of regular maintenance jobs? It is available all the time, since gitlab requires sidekiq.

jvanbaarsen:
@rumpelsepp The way it has to be scheduled is something I'm not a big fan of. Everytime GItLab starts it has to make sure that a job is scheduled in Sidekiq. And what if somehow a job is not rescheduled? You probably will forget about it and it will never happen. Also the creator of Sidekiq is advising against using Sidekiq for recurring jobs: https://github.com/mperham/sidekiq/wiki/Scheduled-Jobs#recurring-jobs

per-garden:
An advantage of using sidekiq is being able to monitor in the admin console, and I think this is an important advantage. I initially attempted scheduling with the whenever gem, without getting it to work properly... It sounds though, that I should pick that up again, and (at the price of losing monitoring) an alternative setup might then be whenever for scheduling and a sidekiq job when gc actually runs. I'll try to look into that...

per-garden:
I still think my original solution is better though... Using whenever:

 1. Nothing happens unless 'whenever -i'
 2. How to tell when gc is scheduled? Logging on to host and 'crontab -l' as opposed to checking in admin GUI?

randx:
OK first of all implementation with `sidetiq`  was ok. We use it for scheduled tasks in EE and its logical to use it in CE too. Also using `sidetiq` requires no additional implementation from package system. So no `whenever` please. 

Second question -  what are advantages of scheduling `git gc` on all repos?

per-garden:
1. My initial implementation does without both sidetiq and whenever. It's self-rescheduling sidekiq job.

2. By issuing "git gc --auto" on all repos, I let git itself determine whether garbage collection is needed, and no extra logic (e.g. checking repo disk usage) is needed in Gitlab itself.

randx:
> My initial implementation does without both sidetiq and whenever. It's self-rescheduling sidekiq job.

@per-garden all repeating jobs should be done via `sidetiq` for consistency with EE. 

> By issuing "git gc --auto" on all repos, I let git itself determine whether garbage collection is needed, 

@per-garden I thought it is done automatically during pushes. 
 
@jacobvosmaer do you think it makes sense to run  `git gc --auto` on all repos on regular schedule?

jacobvosmaer:
Having one job that loops through all repos is a bad idea on large sites like gitlab.com. That would be the never-ending cron (??) job.

@jvanbaarsen I prefer sidetiq (note the `t`) to cron because cron adds an extra dependency and because it is the wrong tool to use if you have a gitlab cluster: you want the work to be done once for the whole cluster, not once per server in the cluster.

Doing this right, in a way that scales, may get very complicated. Do we even need this? Won't Git run `git gc` during certain operations, e.g. on push?

jacobvosmaer:
Example of a complicated solution that might scale:

- create a Sidetiq GcScheduler worker that loops through all projects and makes sure they have a GC scheduled. If not GC is scheduled, schedule it at a _random_ time in the coming week/month
- create a Sidekiq GC worker that does GC on one project (repo + satellite?)

If `git gc` happens anyway on push this seems like a lot of work to have those gc's happen in the background to me.

dosire:
@jacobvosmaer Having a scalable GC solution is hard. I think the scheduling worker is interesting. We only need to GC the repo (satellite will be deprecated). Maybe we can also consider launching a gc run every 500 pushes or so.

jacobvosmaer:
> Maybe we can also consider launching a gc run every 500 pushes or so.

This is default Git behavior, I _think_. (see `--auto` under https://www.kernel.org/pub/software/scm/git/docs/git-gc.html#_options)

I can think of two reasons to do proactive GC:
- detect silent corruption of 'forgotten' repos
- prevent users from having to wait for a repack on the server in day to day use

I am not sure if 'wait for a repack on the server' really is a problem or what its impact is.

dosire:
@jacobvosmaer if git does it automatically that is good enough for me for now

GLundh:
@dosire: Previous to Git v2.0, receive-pack's auto gc holds the push operation hostage until the gc completes. Depending on your repositories and hardware, this can take a long time. And the user will certainly believe something went wrong when the push hangs for 10 minutes+. So if you are running an active server, you really have to disable auto gc on receive-packs. 

However, since Git v2.0, the auto gc will detach and release the push, while it continues the repack. Which is great.

But for the poor servers running Git <v2.0 this is a real issue. Proactive offline gc:ing is almost a must. And since Gitlab supports Git v1.7.10, I think it makes sense to allow this kind of gc scheduling and even include the functionality in GitLab to promote regularly scheduled "offline" GC:ing.

randx:
@GLundh can we just update git version on servers instead? Why try to solve problem which is already solved with new version of git?

jacobvosmaer:
Thanks for explaining @GLundh !

For what it's worth, the omnibus-gitlab packages ship with Git 2.4.3 [at the moment](https://gitlab.com/gitlab-org/omnibus-gitlab/blob/7-12-stable/config/software/git.rb#L20).

GLundh:
@randx: That is a possibility. But with the semi-recent bitmap index support in Git repacks, which costs lots of memory during generation for gits with extreme object counts + the usual high CPU and disk load that occurs during GCing of large gits, many organisations wants to ensure the GC:ing occurs during non peak-times. So I still see a need for it, at least for serious installations.

randx:
@GLundh I am still not sure it worth implementing `gc` over all repos on schedule just to make sure one big repo out of hundred get `gc` not in peek time.

dosire:
Git runs git gc --auto automatically on push, and in recent versions this is asynchronous I now see.

PR #9327, opened 2015-05-24
Possibility to query ldap user on realtime.

farchanjo:
Sometimes we need a realtime query on ldap. LDAP was developed to receive to many read queries per seconds. 1 hour of query is too much time if one bad developer.  As example my ldap filter:

user_filter: '(&(&(VCS=1)(disabled=0))(objectClass=SkoobServices))'

If I set VCS (Version Control System = 0 or disabled = 1), I still can commit via SSH protocol due the cache connection.

I gave the possibility to cache if wanted.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -500,7 +500,11 @@ class User < ActiveRecord::Base
     if !Gitlab.config.ldap.enabled
       false
     elsif ldap_user?
-      !last_credential_check_at || (last_credential_check_at + 1.hour) < Time.now
+      if Gitlab.config.ldap.servers.main.check_cache
+         !last_credential_check_at || (last_credential_check_at + 1.hour) < Time.now
+      else
+         true
+      end
     else
       false
     end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -500,11 +500,7 @@ class User < ActiveRecord::Base
     if !Gitlab.config.ldap.enabled
       false
     elsif ldap_user?
-      if Gitlab.config.ldap.servers.main.check_cache
-         !last_credential_check_at || (last_credential_check_at + 1.hour) < Time.now
-      else
-         true
-      end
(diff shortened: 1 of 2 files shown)


jvanbaarsen:
@farchanjo Can you give a better use case for this? "1 hour of query is too much time if one bad developer" That sounds like something that the developer itself need to change and not something GitLab needs to take in account right?

farchanjo:
I will give better example. I want to disable my developer from my ldap structure and this developer has malicious intentions. One hour as SSH Protocol can erase all project from git. That's the point. 



Atenciosamente,


Fabricio Archanjo Fonseca

jvanbaarsen:
ah ok, now I see. Since we are trying to move a lot of configuration options to the admin UI, would it be possible that you move this as well there?

farchanjo:
I think there are more sense keep on gitlab.yml, because all ldap setting still there. Don’t you think? 

Atenciosamente,


Fabricio Archanjo Fonseca

jvanbaarsen:
@randx What do you think about where to put the settings? And the feature in general?

farchanjo:
I don't think that I do can be the better option. May I suggest another kind the option in another commit?

farchanjo:
The second commit is smarter then the first commit and simple.

randx:
@jvanbaarsen i think we should put all new settings in UI. Feature makes sense for me

farchanjo:
I think does not make sense out this feature on Admin UI if all setting for ldap still on file.



Atenciosamente,


Fabricio Archanjo Fonseca

farchanjo:
I think does not make sense put this feature on Admin UI if all setting for ldap still on file.

jvanbaarsen:
@farchanjo We will a lot of ldap functionality to the admin UI eventually.

farchanjo:
Make sense now for me.

jvanbaarsen:

PR #9335, opened 2015-05-27
[Documentation] non-omnibus instructions for changing time zone

ghost:
With this PR I added some clarifications for changing the time zone when using a non-omnibus installation (time zone needs to be changed in `config/gitlab.yml` and so on).

I moved the "old" omnibus specific instructions to a new section.

So users with a non-omnibus installation can easily adjust the GitLab time zone :)


jvanbaarsen:
Thanks

PR #9340, opened 2015-05-28
Update noteable after a new note is added

jvanbaarsen:
**What does this do?**
It makes sure that whenever a new note is added to a `noteable` item, the
`updated_at` of that item is also updated.

**Why is this needed?**
At this moment when you post a comment on an issue or add a label to an issue,
the `updated_at` is not changed. Because of this the filtering for least recently
updated is not really useful (since it only takes in account the original text
from the noteable).

### DIFF
--- a/app/models/note.rb
+++ b/app/models/note.rb
@@ -31,7 +31,7 @@ class Note < ActiveRecord::Base
   participant :author, :mentioned_users
 
   belongs_to :project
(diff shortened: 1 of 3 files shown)


jvanbaarsen:
Hm ok, I need to fix a broken test

tsigo:
:+1: Good idea.

randx:
I will merge after `7-12-stable` created

PR #9354, opened 2015-06-03
Fix group remove [For test purposes]

dzaporozhets:
Same PR in github has valid diff

cc @DouweM

### DIFF
--- a/app/models/namespace.rb
+++ b/app/models/namespace.rb
@@ -72,7 +72,7 @@ class Namespace < ActiveRecord::Base
       path.gsub!(/[^a-zA-Z0-9_\-\.]/, "")
 
       # Users with the great usernames of "." or ".." would end up with a blank username.
-      # Work around that by setting their username to "blank", followed by a counter. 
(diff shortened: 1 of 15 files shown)


app/models/namespace.rb
```diff
@@ -99,7 +99,18 @@ def ensure_dir_exist
...
+    # We will remove it later async
+    new_path = "#{path}+#{id}+deleted"
+
+    if gitlab_shell.mv_namespace(path, new_path)
+      message = "Namespace directory \"#{path}\" moved to \"#{new_path}\""
+      Gitlab::AppLogger.info message
+
+      # Remove namespace directroy async with delay so
+      # GitLab has time to remove all projects first
+      GitlabShellWorker.perform_in(5.minutes, :rm_namespace, new_path)
```
randx: @jacobvosmaer I made similar approach for namespace in case when we remove user or group

jacobvosmaer: Cool!

PR #9368, opened 2015-06-08
No need to check if `repository_ref` is present

ghost:
There is no need to check if `repository_ref` is present as:

```
@repository_ref = if repository_ref.present?
                    repository_ref
                  else
                    nil
                  end
```

is as same as doing:

```
@repository_ref = repository_ref
```

### DIFF
--- a/lib/gitlab/project_search_results.rb
+++ b/lib/gitlab/project_search_results.rb
@@ -4,11 +4,7 @@ module Gitlab
 
     def initialize(project_id, query, repository_ref = nil)
       @project = Project.find(project_id)
-      @repository_ref = if repository_ref.present?
-                          repository_ref
-                        else
-                          nil
-                        end
+      @repository_ref = repository_ref
       @query = Shellwords.shellescape(query) if query.present?
     end
 


zenati:
This MR has been transfered to https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/784.

stanhu:
Is that true? If `repository_ref` is `""`, then `@repository_ref` will be `nil` in the original case, whereas it will be `""` in the new case. It may be worthwhile to check how it's being used.

DouweM:
@stanhu is right. There's also most likely a reason why the check was added. 

@zenati please double check that this doesn't break (or otherwise change) expected behavior.

jvanbaarsen:
@DouweM Hm ok, overlooked this (Also build was green). Shall I revert? Or wait for @zenati to double check?

DouweM:
@jvanbaarsen Revert, please. The repository_ref seems to originate in a param, so it could very well be an empty string, in which case we want to treat it as nil.

Razer6:
Maybe add a regression test to catch this behaviour.

jvanbaarsen:
@Razer6 :+1: for regression test

stanhu:
MR for the test here: https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/788

PR #9371, opened 2015-06-09
An `in_namespace` scope is already present

ghost:
An `in_namespace` scope is already present and never used.
See: https://github.com/gitlabhq/gitlabhq/blob/master/app/models/project.rb#L161

Instead of always using `where(namespace_id: VALUE)`, using the scope `in_namespace` would be appropriate.


randx:
:+1:

PR #9373, opened 2015-06-09
Revert "No need to check if `repository_ref` is present"

jvanbaarsen:
Reverts gitlabhq/gitlabhq#9368 because of not explicit nil check


DouweM:
@jvanbaarsen <3

PR #9375, opened 2015-06-10
Add satellite create command for omnibus installs

donoftime:
In at least the 7.9.1 omnibus release and after I was able to use
the `gitlab-rake` command to rebuild the satellites, rather than
using bundle directly. This doc entry brings the satellite section
into alignment with the other sections in the doc which show both
the use of gitlab-rake and the direct execution of the rake task
via bundle.

ci-skip

### DIFF
--- a/doc/raketasks/maintenance.md
+++ b/doc/raketasks/maintenance.md
@@ -117,6 +117,12 @@ This will create satellite repositories for all your projects.
 
 If necessary, remove the `repo_satellites` directory and rerun the commands below.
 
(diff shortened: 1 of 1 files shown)


axilleas:
Thanks for the contribution but satelites will be dropped somewhere between the next two releases, as you can see from this MR https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/981

donoftime:
Cool, that is even better.

PR #9384, opened 2015-06-11
No need to check twice for `nil?` as `blank?` already does it

ghost:
No need to check twice for `nil?` as `blank?` already does it. Because `'', ' ', nil, [], and {}` are all blank.

See: http://api.rubyonrails.org/classes/Object.html#method-i-blank-3F

```
return if params[:search].nil? || params[:search].blank?
```

Should be:

```
return if params[:search].blank?
```

### DIFF
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -4,7 +4,7 @@ class SearchController < ApplicationController
   layout 'search'
 
   def show
-    return if params[:search].nil? || params[:search].blank?
+    return if params[:search].blank?
 
     @search_term = params[:search]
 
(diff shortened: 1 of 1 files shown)


stanhu:
Thanks--can you squash this into one commit?

app/controllers/search_controller.rb
```diff
@@ -4,7 +4,8 @@ class SearchController < ApplicationController
   layout 'search'
 
   def show
-    return if params[:search].nil? || params[:search].blank?
+    # See: http://api.rubyonrails.org/classes/Object.html#method-i-blank-3F
```
randx: I think this comment is ambiguous. Following this pattern we can include URL with explanations to thousand of lines of code.

PR #9400, opened 2015-06-19
Use whoami instead of $USER

tuxillo:
- Use whoami instead of relying on the existence of $USER enviroment variable which is not always present


axilleas:
ping @jacobvosmaer

jacobvosmaer:
Thanks @tuxillo

PR #9411, opened 2015-06-24
Fixed problem with non-existing user not being detected.

ghaering:
The effect I saw in production was that a "random" user was being shown in the activity log instead of the correct one. I think I found the reason for this. The change was not tested, and I do not really know Ruby but I hope I got the change from member to local variable right.

"@user" being a member variable, there normally was a @user object active, and ||= would keep it around if identify returns nil. Changing @user to a local variable fixes that and now non-existing users will actually be detected.

### DIFF
--- a/app/workers/post_receive.rb
+++ b/app/workers/post_receive.rb
@@ -28,17 +28,17 @@ class PostReceive
     changes.each do |change|
       oldrev, newrev, ref = change.strip.split(' ')
 
-      @user ||= identify(identifier, project, newrev)
+      user ||= identify(identifier, project, newrev)
 
-      unless @user
+      unless user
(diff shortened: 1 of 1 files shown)


stanhu:
Thanks for the pull request. I think the changes do make sense. Was `@user` copied from a controller, @randx?

jvanbaarsen:
@stanhu I don't think `@user` can be from a controller, since this is a worker, so it doesnt have access to the controller code because different threads. But code looks good to me. @randx Would you be conformable with merging without tests?

randx:
With this patch `user` variable will be set every time and `||=` is useless. Push is always done by single person. So inside push it should be 1 author always

jvanbaarsen:

PR #9418, opened 2015-06-28
Add a project service for LetsChat including tests

arifumi:
[Lets Chat](http://sdelements.github.io/lets-chat/) is a open-source self hosted chat software.
This patch adds support for Lets Chat as a project service.

I've almost followed the manner the HipChat service does, including spinach and rspec test code. But, it does not include API and documentation parts yet.


stanhu:
LetsChat looks like it is trying to be an open source version of Slack. Looks promising. I'm all for this change, although I wonder if there's a way we can avoid this much code duplication between HipChat.

jvanbaarsen:

jvanbaarsen:
This PR has been closed due to inactivity

PR #9438, opened 2015-07-07
dont -> don't

bbodenmiller:
Update "You dont have any authorized applications" to "You don't have any authorized applications"


stanhu:
Thanks. There are other sources where "dont" should be changed to "don't. For example:

https://github.com/gitlabhq/gitlabhq/blob/ce5928f4df1ebdac1fbc319fd1dbd15585f5da9e/app/services/delete_branch_service.rb#L24

I think the other cases are just comments and aren't user visible at the moment.

https://github.com/gitlabhq/gitlabhq/search?utf8=%E2%9C%93&q=dont

PR #9448, opened 2015-07-11
Fix problem with API Transfer Project

sazor:
I had issues while trying to transfer project to group through api. So I found out that method execute use new_namespace_id param instead of namespace_id. But in API method's called with namespace_id param.


stanhu:
Thanks for your contribution. Would you mind adding a test in `spec/requests/api/groups.rb` so that this can be verified?

sazor:
This commit is not needed because of this fix https://github.com/gitlabhq/gitlabhq/commit/b93053695a94822801a87ff2b1850999130842c1

PR #9455, opened 2015-07-13
Remove author information from "post" in merge_requests API.

rburchell:
From both experimenting with the API and reading the source code, this appears to not be used at all.

My guess is this was copypasta from somewhere else.


stanhu:
Looks like you're right. Thanks, @rburchell!

PR #9456, opened 2015-07-13
Update password in README.md

PiotrCzapla:
It seems that the password was changed see
https://about.gitlab.com/downloads/

At least the updated password works for me, while the old "password" wasn't


stanhu:
We're in a transition period where the new password hasn't actually been released in GitLab but will be in v7.13. Since this has come up before, perhaps it may be helpful to document both old and new passwords depending on the version of GitLab?

cc: @tsigo

tsigo:
The password was changed (to `password`) a while ago and then changed back today. It will remain `5iveL!fe` for the time being.

Thanks!

PR #9469, opened 2015-07-21
Fix Issue 707 (in gitlab): Indent unfolded code 1 character

Leibniz137:
This is a fix for https://gitlab.com/gitlab-org/gitlab-ce/issues/707

Note that if passed an invalid value is passed as the 'input' parameter (e.g. -1), Gitlab will 500. Not sure if this is acceptable behavior.

Manually tested GET'ing diffs with a valid value passed as the 'indent' parameter. Also tested GET'ing a diff without specifying the 'indent' param, confirmed that no indentation was added.

Before unfold:
screen shot 2015-07-21 at 11 23 17 am

After unfold:
screen shot 2015-07-21 at 11 21 50 am

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,7 +1,8 @@
 Please view this file on the master branch, on stable branches it's out of date.
 
 v 7.14.0 (unreleased)
-  - Remove repository graph log to fix slow cache updates after push event (Stan Hu)
+  - Fix full screen mode for snippet comments (Daniel Gerhardt)
+  - Fix 404 error in files view after deleting the last file in a repository (Stan Hu)
   - Fix label read access for unauthenticated users (Daniel Gerhardt)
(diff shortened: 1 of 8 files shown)


app/views/projects/blob/diff.html.haml
```diff
@@ -11,7 +11,7 @@
       %td.old_line.diff-line-num{data: {linenumber: line_old}}
         = link_to raw(line_old), "#"
       %td.new_line= link_to raw(line_new) , "#"
-      %td.line_content.noteable_line= line
+      %td.line_content.noteable_line= ' '*@form.indent + line
```
stanhu: Can you add a spaces around the operator `*`?

stanhu:
Good catch. In general looks good. Would you mind squashing this into one commit?

Issue #9465, opened 2015-07-20
Broken GitLab 7.12 service


NuLL3rr0r:
Hello today I did an upgrade from 7.11.4 to 7.12.2. When I did:
service gitlab start
I faced with a new shell instead of starting GitLab. In the new shell running whoami says it is git user.

I did a diff:
+shell_path="/bin/bash"
 # Switch to the app_user if it is not he/she who is running the script.
 if [ "$USER" != "$app_user" ]; then
-  eval su - "$app_user" -c $(echo \")$0 "$@"$(echo \"); exit;
+  eval su - "$app_user" -s $shell_path -c $(echo \")$0 "$@"$(echo \"); exit;

And found two changes:
shell_path="/bin/bash"
On FreeBSD systems there is no Bash by default and when you install it either from Ports or pkgng it resides in /usr/local/bin/bash. So I changed it to its correct value on FreeBSD. But, I believe the following is a cross platform way of accessing a shell:
/usr/bin/env ${SHELL_NAME}, e.g.: /usr/bin/env bash

It did not solve the problem anyway, and the culprit was the second change. reverting to
eval su - "$app_user" -c $(echo \")$0 "$@"$(echo \"); exit;
(without shell_path variable) solved the issue for me.

JacquesTheRock:
What about setting shell_path to $SHELL or possibly $(which $SHELL)?

NuLL3rr0r:
Yep thanks @JacquesTheRock

/usr/bin/env ${SHELL_NAME}, e.g.: /usr/bin/env bash
Will drop you inside the shell you want (e.g. bash) probably it's not a solution here. I realized my error.

Anyway, your solution did the trick.
shell_path="$SHELL"
works just fine. Thank you so much.
And if the authors still need bash it should be:
shell_path="$(which bash)"

Thanks, I'll create a pull request for that. But, there's still one annoying issue when I do "service gitlab restart" after starting the service I'll end up in a shell which is owned by GitLab user (on my system 'git').
So, I should type 'exit' to get back to root shell.

PR #9477, opened 2015-07-23
Fix for issue #9465

NuLL3rr0r:
$SHELL returns the correct shell path on *nix system

### DIFF
--- a/lib/support/init.d/gitlab
+++ b/lib/support/init.d/gitlab
@@ -35,7 +35,7 @@ pid_path="$app_root/tmp/pids"
 socket_path="$app_root/tmp/sockets"
 web_server_pid_path="$pid_path/unicorn.pid"
 sidekiq_pid_path="$pid_path/sidekiq.pid"
-shell_path="/bin/bash"
+shell_path="$SHELL"
 
 # Read configuration variable file if it is present
 test -f /etc/default/gitlab && . /etc/default/gitlab


stanhu:
Seems reasonable to me. @jacobvosmaer, any reason this shouldn't go in?

jacobvosmaer:
$SHELL returns the shell of the current user. It is not uncommon for root to use a different shell than normal users, and commands like `su` and `sudo` have complicated rules/options for either changing or not changing $SHELL when switching user. I don't think relying on $SHELL in an init script is a good idea.

I know that systems like FreeBSD do not have /bin/bash, even if you install bash it goes into another directory.

If `shell_dir=/bin/bash` is a problem on some systems, I think the solution should be to make the init script work with `sh` (plain POSIX shell) rather than relying on Bash.

TL;DR use /etc/default/gitlab if your bash is not in /bin. The 'real' solution is to use /bin/sh in the init script instead of Bash.

jacobvosmaer:
You can just write this in /etc/default/gitlab:

```
shell_path=/usr/local/bin/bash
```

NuLL3rr0r:
Using /bin/sh seems a good idea to me. But there is no /etc/default/gitlab on my system since I did installed it from source. And I never saw anything about it on the installation docs. But I'll check it out.
By the way on *BSD operating system anything outside base system goes into /usr/local/ e.g. /usr/local/bin for binaries and /usr/local/etc/ for configuration files or /usr/local/etc/rc.d/ for init scripts. So I suppose it must be /usr/local/etc/default/gitlab on BSD family of operating systems.

jacobvosmaer:
@NuLL3rr0r see this comment for the /etc/default/gitlab file: https://github.com/gitlabhq/gitlabhq/blob/82f0d3e655c48fc67adea88af681d0b5d81810de/lib/support/init.d/gitlab#L20

There is also a template in https://github.com/gitlabhq/gitlabhq/blob/master/lib/support/init.d/gitlab.default.example

jacobvosmaer:
> So I suppose it must be /usr/local/etc/default/gitlab on BSD family of operating systems.

Yes something like that would be more natural I agree. But I am not sure how to fix that in the init script.

Wouldn't you want a completely different init script that just uses rc.subr anyway? That would be much nicer on FreeBSD now that I think about it.

NuLL3rr0r:
Sorry, my mistake. I thought /etc/default/gitlab is a configuration file that you are mentioning. Yes I already have that file. In fact as part of my installation or upgrade process I always do:
$ sudo cp lib/support/init.d/gitlab /usr/local/etc/rc.d/gitlab

> You can just write this in /etc/default/gitlab:
>    shell_path=/usr/local/bin/bash

Yes, the first thing I did was that. But, until now (GitLab < 7.12.0) I enjoyed installing it like Linux users and it worked out of the box.

> Wouldn't you want a completely different init script that just uses rc.subr anyway? That would be much nicer on FreeBSD now that I think about it.

I never thought of that. So, if it's OK with you I can write one in my spare time since the current one is well documented and contribute it back to be merged.

jacobvosmaer:
This somehow got snowed under in my inbox.

Having had to touch the init script again recently, I think life could be a lot easier on FreeBSD with a native version. It is not something I can actively work on however.

At this point GitLab consists of 4 daemons: unicorn, sidekiq, mail_room and gitlab-git-http-server. I still do not really understand what is the canonical way to manage such a group in FreeBSD, or any other init system for that matter.

NuLL3rr0r:
OK, then I start working on it. But I won't promise a deadline since I'll try working on it on my spare time. I'll report back if I have any success with it.

jvanbaarsen:

PR #9480, opened 2015-07-24
Prevent personal pronouns.

eliasp:


stanhu:
I'm not sure how I feel about this one. @dosire, any opinions? :)

dosire:
In CONTRIBUTING we have:
Interface text should be written subjectively instead of objectively. It should be the gitlab core team addressing a person. It should be written in present time and never use past tense (has been/was). For example instead of "prohibited this user from being saved due to the following errors:" the text should be "sorry, we could not create your account because:". Also these excellent writing guidelines https://design.atlassian.com/latest/product/foundations/writing-style/

It seems that using you is encouraged in these guides.

@eliasp What do you think?

eliasp:
You're right - didn't see this guideline. I was used to UX concepts where "the software" doesn't address "the user" personally.
The status quo is in line with your project's guidelines, so it's fine to keep it that way.

dosire:
Thanks @eliasp

PR #9490, opened 2015-07-25
clone only last revision

manfer:
We don't need to have the full branch history just to install an app from source. The last revision is more than enough.

If you clone the whole branch you need more than 80MB. If you clone only last revision you need less than 7MB.

Not only you are saving your disk space, you are saving github bandwidth and you are saving your time not having to wait for 10 times more MB to download.


stanhu:
I can see this being useful, but I'm not sure if the average user cloning from source will only want the last revision. It might be better left as a tip for users who want to conserve disk space and bandwidth.

PR #9497, opened 2015-07-28
Update projects_controller.rb

mikicaivosevic:
Refactor show method

PR #9498, opened 2015-07-28
Update projects_controller.rb

mikicaivosevic:
Refactor show method if else


stanhu:
Seems to be a duplicate of #9497.

PR #9504, opened 2015-07-29
Added space between words

mr-dxdy:


stanhu:
Doesn't HAML just filter this out? Do you need to use [`surround`](http://haml.info/docs/yardoc/Haml/Helpers.html#surround-instance_method)?

mr-dxdy:
@stanhu 
email-image

stanhu:
What browser are you using? I'm not seeing this on Chrome:

image


Also, my point was that I thought HAML stripped the trailing and leading whitespace around the text, so adding a space at the end may not do anything. Or does it?

mr-dxdy:
@stanhu  i see it on browsers Chrome 44.0.2403.125 and Firefox 39.0. 
I checked the other letters and the letter from Github incorrectly displayed:

github-email 
 
I have no idea where is the problem. I close pull request...

PR #9521, opened 2015-08-02
@filter_params does not have to be an instance variable, so it was deleted.

gipcompany:
@filter_params does not have to be an instance variable, so it was deleted.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -269,32 +269,30 @@ class ApplicationController < ActionController::Base
     params[:state] = 'opened' if params[:state].blank?
 
     @sort = params[:sort]
-    @filter_params = params.dup
+    filter_params = params.dup
 
     if @project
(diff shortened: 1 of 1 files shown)


stanhu:
Should `set_filter_params` be renamed to `filter_params` then?

gipcompany:
@filter_params is accessible by others(controllers and views) and might have some possibilities to be changed, but I confirmed nothing accesses the instance variable. Also, I don't know the reason why that parameter should be used as an instance variable. So, in that context you don't have to use an instance variable, you should just use set_filter_params method instead. The tiny refactoring should be thought better for code security and maintenance(by less code).

stanhu:
I understand the change, but if `set_filter_params` is no longer setting an instance variable, it seems like it should be renamed.

gipcompany:
+1, I agree to the rename as you said. But I don't have an appropriate name idea so far....

jvanbaarsen:

PR #9526, opened 2015-08-04
fix bug 'ActionView::Template::Error (uninitialized constant Gitlab::…

t-zuehlsdorff:
fix bug 'ActionView::Template::Error (uninitialized constant Gitlab::Markdown::AutolinkFilter::Rinku):'

This bug was found when porting GitLab to FreeBSD. Its not possible to display Readme.md or some service templates - Jira for example - without this patch.


rumpelsepp:
`gemfile.lock` changes are missing

t-zuehlsdorff:
I see, BUT it is not missing. Rinku is already in Gemfile.lock. But not in Gemfile itself!
Also it is registered as dependency of wikicloth, but i could not find it there in the Gemfile nor gemspec.

stanhu:
This test is failing at the moment. You have to run `bundle install` and then check in `Gemfile.lock`.

t-zuehlsdorff:
As already statet the testcase is wrong, not me ;)
But after some updates i could not update this pull request. Please use the new one instead:
https://github.com/gitlabhq/gitlabhq/pull/9609

PR #9535, opened 2015-08-06
check upload dir permissions

bbodenmiller:
Detect permission and ownership issues with the users uploads dir. This helps troubleshoot and correct [gitlab-org/omnibus-gitlab#311](https://gitlab.com/gitlab-org/omnibus-gitlab/issues/311). Fixes gitlabhq#7500, gitlabhq#7052. Related to gitlabhq#6281. Replaces #8769.

### DIFF
--- a/lib/tasks/gitlab/check.rake
+++ b/lib/tasks/gitlab/check.rake
@@ -22,6 +22,7 @@ namespace :gitlab do
       check_gitlab_config_not_outdated
       check_log_writable
       check_tmp_writable
+      check_uploads
       check_init_script_exists
       check_init_script_up_to_date
       check_projects_have_namespace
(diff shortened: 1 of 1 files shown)


lib/tasks/gitlab/check.rake
```diff
@@ -276,6 +277,45 @@ namespace :gitlab do
...
+          return
+        end
+
+        # if tmp upload dir has incorrect permissions, assume others do as well
+        if File.stat(upload_path_tmp).mode == 040755 && File.owned?(upload_path_tmp) # verify drwxr-xr-x permissions
+          puts "yes".green
+        else
+          puts "no".red
+          try_fixing_it(
+            "sudo chown -R #{gitlab_user} #{upload_path}",
```
stanhu: Does the group name also need to be specified (e.g. `gitlab-www`)? See: https://gitlab.com/gitlab-org/omnibus-gitlab/issues/254

lib/tasks/gitlab/check.rake
```diff
@@ -276,6 +277,45 @@ namespace :gitlab do
         fix_and_rerun
       end
     end
+    
+    def check_upload_perms
+      print "Upload directory permissions correct? ... "
+
+      upload_path = File.realpath(Rails.root.join('public/uploads'))
```
stanhu: `realpath` will throw an exception if the directory does not exist.

lib/tasks/gitlab/check.rake
bbodenmiller: Non-Omnibus GitLab upload path does not have a group and Omnibus GitLab fixes the group with `gitlab-ctl reconfigure` per https://github.com/gitlabhq/gitlabhq/pull/8769#discussion_r24588605 so no need to check and fix it here. Also per https://gitlab.com/gitlab-org/omnibus-gitlab/issues/311#note_348792 you wouldn't want to run a recursive applying of the group anyways.

lib/tasks/gitlab/check.rake
bbodenmiller: Suggestions for a fix?

stanhu: You can check that it exists and is a directory first via `File.directory?`.

lib/tasks/gitlab/check.rake
stanhu: Ok, sounds good.

lib/tasks/gitlab/check.rake
bbodenmiller: Fixed in latest commit.

stanhu:
@jacobvosmaer, can you review this one?

lib/tasks/gitlab/check.rake
jacobvosmaer: Yeah everything under public/uploads belongs to 'git'. The git:gitlab-www ownership on public/uploads itself is just to give the gitlab-www group read(-only) access.

lib/tasks/gitlab/check.rake
```diff
@@ -276,6 +277,57 @@ namespace :gitlab do
...
+          )
+          for_more_information(
+            see_installation_guide_section "GitLab"
+          )
+          fix_and_rerun
+        end
+      else
+        puts "no".red
+        try_fixing_it(
+          "sudo chmod 0750 #{upload_path}",
```
jacobvosmaer: Where in the big if-else-if-else are we here? Is the file mode the only thing that's wrong in this code branch, or could the owner also be wrong?

jacobvosmaer:
@bbodenmiller looks nice

@stanhu what is the thing about long argument lists you talk about [above](#commitcomment-12741005)? I don't see it in the code.

stanhu:
Never mind; looks like the `-exec command {}` argument sends one command per filename; I thought `find` was bundling the command-line into one long string.

Looks good to me too.

lib/tasks/gitlab/check.rake
bbodenmiller: Are you proposing changing

`if File.stat(upload_path).mode == 040750`

to

`if File.stat(upload_path).mode == 040750 && File.owned?(upload_path)`?

axilleas:
@jacobvosmaer @stanhu can we merge this then?

jacobvosmaer:
Thanks @bbodenmiller !

PR #9538, opened 2015-08-07
Fogbugz importer

szechyjs:
This is nearly feature complete, but before I put the finishing touches on it I wanted to open a discussion and make sure I'm heading down the right path.

Since Fogbugz does not directly couple their repositories (Kiln) with their bug reporting this importer just creates a bare repository and then populates the issues, comments, and labels.

One thing that I would like to do as well is have it import attachments. I'm not sure the best route for going about this. During the import I have access to a URL of the attachment, this URL requires authentication/token with Fogbugz, so it would make sense to upload it as an attachment to gitlab.


axilleas:
\cc @DouweM

szechyjs:
@axilleas Not a problem, I'll definitely add some documentation.

Procfile
```diff
@@ -1,2 +1,3 @@
 web: bundle exec unicorn_rails -p ${PORT:="3000"} -E ${RAILS_ENV:="development"} -c ${UNICORN_CONFIG:="config/unicorn.rb"}
 worker: bundle exec sidekiq -q post_receive -q mailer -q archive_repo -q system_hook -q project_web_hook -q gitlab_shell -q common -q default
+log: tail -f log/development.log
```
DouweM: Please remove this :)

app/controllers/import/fogbugz_controller.rb
```diff
@@ -0,0 +1,106 @@
...
+
+  rescue_from Fogbugz::AuthenticationException, with: :fogbugz_unauthorized
+
+  def new
+
+  end
+
+  def callback
+    keyset = [:uri, :email, :password]
+    res = Gitlab::FogbugzImport::Client.new(params.slice(*keyset).symbolize_keys)
```
DouweM: Please create a new method `import_params`:

```ruby
def import_params
  params.permit(:uri, :email, :password)
end
```

And use that here:

```ruby
res = Gitlab::FogbugzImport::Client.new(import_params)
```

app/views/import/fogbugz/new.html.haml
```diff
@@ -0,0 +1,45 @@
...
+        Log into your Fogbugz account and set your filter to show All cases.
+      %p
+        =image_tag 'fogbugz_filter.png'
+    %li
+      %p
+        Enter your Fogbugz URL and login information.
+      %p
+        = label_tag :uri do
+          Fogbugz URL
+          = text_field_tag :uri
```
DouweM: Could you add a `placeholder` with an example? Right now, it isn't clear if this should be just the host or a full path.

app/views/import/fogbugz/new.html.haml
```diff
@@ -0,0 +1,45 @@
...
+= form_tag callback_import_fogbugz_path, class: 'form-horizontal', multipart: true do
+  %p
+    Follow the steps below to export your Fogbugz project data.
+    In the next step, you'll be able to select the projects you want to import.
+  %ol
+    %li
+      %p
+        Log into your Fogbugz account and set your filter to show All cases.
+      %p
+        =image_tag 'fogbugz_filter.png'
```
DouweM: Please move this into an appropriate folder.

DouweM: Also, why is this step necessary if everything is imported using the API?

app/views/import/fogbugz/new_user_map.html.haml
```diff
@@ -0,0 +1,37 @@
...
+  %i.fa.fa-bug
+  Import projects from Fogbugz
+%hr
+
+= form_tag create_user_map_import_fogbugz_path, class: 'form-horizontal' do
+  %p
+    Customize how Fogbugz email addresses and usernames are imported into GitLab.
+    In the next step, you'll be able to select the projects you want to import.
+  %p
+    The user map is a JSON document mapping the Fogbugz users that participated on your projects to the way their email address and usernames wil be imported into GitLab. You can change this by changing the value on the right hand side of <code>:</code>. Be sure to preserve the surrounding double quotes, other punctuation and the email address or useranem on the left hand side.
```
DouweM: "useranem" -> "username"

app/views/import/fogbugz/new_user_map.html.haml
```diff
@@ -0,0 +1,37 @@
...
+      %p
+        <code>"johnsmith@example.com": "@johnsmith"</code>
+        will add "By <a href="#">@johnsmith</a>" to all issues and comments originally created by johnsmith@example.com,
+        and will set <a href="#">@johnsmith</a> as the assignee on all issues originally assigned to johnsmith@example.com.
+    %li
+      %strong Map a Fogbugz user to a full email address
+      %p
+        <code>"johnsmith@example.com": "johnsmith@example.com"</code>
+        will add "By <a href="#">johnsmith@example.com</a>" to all issues and comments originally created by johnsmith@example.com.
+        By default, the email address or username is masked to ensure the user's privacy. Use this option if you want to show the full email address.
```
DouweM: It's not in this case, though?

app/views/import/fogbugz/new.html.haml
```diff
@@ -0,0 +1,45 @@
+- page_title "Fogbugz Import"
+%h3.page-title
+  %i.fa.fa-bug
+  Import projects from Fogbugz
+%hr
+
+= form_tag callback_import_fogbugz_path, class: 'form-horizontal', multipart: true do
+  %p
+    Follow the steps below to export your Fogbugz project data.
```
DouweM: "export" isn't really appropriate here. I'd go with "Follow the steps below to connect to your Fogbugz account."

app/views/import/fogbugz/status.html.haml
```diff
@@ -0,0 +1,70 @@
...
+- if @repos.any?
+  %p.light
+    Select projects you want to import.
+  %p.light
+    Optionally, you can
+    = link_to 'customize', new_user_map_import_fogbugz_path
+    how Fogbugz email addresses and usernames are imported into GitLab.
+  %hr
+  %p
+  - if @incompatible_repos.any?
```
DouweM: If there will never be incompatible repos, we can get rid of these checks.

app/views/projects/new.html.haml
```diff
@@ -66,7 +66,11 @@
 
             = link_to new_import_google_code_path, class: 'btn' do
               %i.fa.fa-google
-              Google Code
+                Google Code
```
DouweM: The label shouldn't be inside the `<i>` tag.

app/workers/repository_import_worker.rb
```diff
@@ -7,10 +7,16 @@ class RepositoryImportWorker
   def perform(project_id)
     project = Project.find(project_id)
 
-    import_result = gitlab_shell.send(:import_repository,
+    if project.import_type != 'fogbugz'
```
DouweM: It's better to check the import_url for `http://unknown.git` here, even better if you put that into a constant on Project.

app/workers/repository_import_worker.rb
```diff
@@ -20,6 +26,8 @@ def perform(project_id)
                             Gitlab::BitbucketImport::Importer.new(project).execute
                           elsif project.import_type == 'google_code'
                             Gitlab::GoogleCodeImport::Importer.new(project).execute
+                          elsif project.import_type == 'fogbugz'
+                            Gitlab::FogbugzImport::Importer.new(project).execute
```
DouweM: This `if/elsif` statement is getting so long that I'd prefer a `case/when` statement. Could you do that?

lib/gitlab/fogbugz_import/client.rb
```diff
@@ -0,0 +1,62 @@
+require 'fogbugz'
+
+module Gitlab
+  module FogbugzImport
+    class Client
+      attr_reader :api
+
+      def initialize(options = {})
+        if options[:uri] && options[:token]
```
DouweM: When is the token option used?

DouweM: Found it :)

DouweM:
@szechyjs Nice work :) Could you post screenshots of the import flow and an imported case/issue here so we can review those as well?

app/views/import/fogbugz/new.html.haml
szechyjs: The fogbugz API only returns the bugs that part of the user's current filter. Unfortunately it seems that through the API there is no way to create a filter, you can only select one.

I'm currently investigating if there is a builtin filter that is the same for all fogbugz users that includes all cases.

szechyjs:
@DouweM thanks for the input, it's exactly what I was looking for.

szechyjs:
Workflow screenshots

new import
user map
select projects
importing
finished
issues
issue

DouweM:
- The log in form on the initial page doesn't look very good.  Try to use `.form-group`, `.control-label`, `.form-control` like elsewhere.
- Mapping user IDs works, but isn't very user friendly. Can't we map usernames?
- I don't think the event description needs the "Description: " prefix.

szechyjs:
I made a bunch of updates.

@DouweM I was able to get in touch with Fog Creek support and they pointed me in the right direction on using the API without setting up the filter through the web interface, of course it wasn't in their documentation.

I also cleaned up the login form, and thanks to the API fix, we don't need the graphic anymore.

There were also a few bugs I encountered that I squashed as well.

Let me know what you think.

Thanks again!

screen shot 2015-08-18 at 7 52 17 pm

DouweM:
@szechyjs Looks good! Did you see my comment "Mapping user IDs works, but isn't very user friendly. Can't we map usernames?"?

szechyjs:
@DouweM Sorry, forgot to mention that in the last comment. What do you suggest for the user_map? I agree when creating the map an ID isn't very friendly.

The fields available to me are:

- User ID
- Full Name
- Email Address

All of the cases reference a user ID, so ultimately I will need to use the ID to lookup the desired name.

stanhu:
Since full name is not necessarily unique (e.g. John Smith) and e-mail addresses may be hard to discern, we may want to show both. I think requiring a user to create a properly-formatted JSON entry could be challenging.

I propose changing the interface so that each mapping consists of a big table with each row containing a separate column for full name and e-mail address. Another column would have a autocomplete selection for a GitLab user.

szechyjs:
@stanhu I like the idea. I'll see what I can do.

szechyjs:
Some new screenshots of the reworked user mapping.

 
screen shot 2015-08-19 at 2 22 45 am
user_map
select_user

lib/gitlab/fogbugz_import/importer.rb
```diff
@@ -0,0 +1,287 @@
...
+
+        @known_labels = Set.new
+      end
+
+      def execute
+        return true unless repo.valid?
+
+        data = project.import_data.try(:data)
+
+        client = Gitlab::FogbugzImport::Client.new(token: data['fb_session']['token'], uri: data['fb_session']['uri'])
```
stanhu: Should some error message be displayed if the user specifies the wrong token and this throws `AuthenticationException` (https://github.com/firmafon/ruby-fogbugz/blob/master/lib/ruby_fogbugz/interface.rb#L11)?

stanhu: Oh, I see the pull request mentioned above that may prevent this from being rescued.

app/views/import/fogbugz/new_user_map.html.haml
```diff
@@ -0,0 +1,45 @@
...
+  %p
+    Customize how FogBugz email addresses and usernames are imported into GitLab.
+    In the next step, you'll be able to select the projects you want to import.
+  %p
+    The user map is a mapping of the FogBugz users that participated on your projects to the way their email address and usernames wil be imported into GitLab. You can change this by populating the table below.
+  %ul
+    %li
+      %strong Default: Map a FogBugz account ID to a full name
+      %p
+        An empty GitLab User field will add "By John Smith" to all issues and comments.
```
stanhu: Is it really "John Smith" or the project creator? :)

stanhu:
Great work! One question about attachments: do the URLs given back by the API require authentication to a Fogbugz server? Or can these URLs can be safely linked directly from GitLab?

CHANGELOG
```diff
@@ -65,6 +65,7 @@ v 7.14.0 (unreleased)
   - Update Flowdock integration to support new Flowdock API (Boyan Tabakov)
   - Remove author from files view (Sven Strickroth)
   - Fix infinite loop when SAML was incorrectly configured.
+  - Add FogBugz project import (Jared Szechy)
```
stanhu: We'll have to bump this to 8.0 now, since 7.14 is feature frozen at the moment.

app/views/import/fogbugz/new_user_map.html.haml
```diff
@@ -0,0 +1,45 @@
...
+
+  %table.table
+    %thead
+      %tr
+        %th ID
+        %th Name
+        %th Email
+        %th GitLab User
+    %tbody
+      - @user_map.each do |id, user|
```
stanhu: Would it help to sort by id or name?

CHANGELOG
axilleas: I created 8.0 milestone and assigned this PR to it.

axilleas:
@szechyjs great work indeed! Let me know if you need any help with the docs ;)

lib/gitlab/fogbugz_import/importer.rb
szechyjs: Yea, I've been in contact with the gem maintainer, we are working on a solution to get a new maintainer.

szechyjs:
@stanhu Right now there is still a little work that needs done on the attachments. The URL returned by the API requires authentication. The API token can be added to the URL to access the attachment. Since most of us will be importing to GitLab so we can get rid of FogBugz, I'd like to come up with a way to essentially import the attachments into GitLab.

The server essentially would have to append the auth token to the URL, download it, 'upload' it to GitLab and then use the new GitLab URL in the Issues.

szechyjs:
It looks like carrierwave supports this type of functionality.

https://github.com/carrierwaveuploader/carrierwave#uploading-files-from-a-remote-location

DouweM:
@szechyjs To upload the file, you should be able to use the `Projects::UploadService`.

The user map UI looks great. When this is done, it'd be awesome if you changed the Google Code one to work like that as well ;)

app/views/import/fogbugz/new_user_map.html.haml
szechyjs: The API is sorting by name by default.

szechyjs:
@DouweM thanks! I have implemented the downloading/saving of attachments.

stanhu:
Looks great so far. Any chance you can add some specs for `Importer`?

szechyjs:
@stanhu specs added

stanhu:
@szechyjs Thanks! Can you rebase master and resolve current conflicts?

szechyjs:
rebased. once we get ready to merge i'll squash as well.

app/services/projects/download_service.rb
```diff
@@ -0,0 +1,40 @@
...
+    def valid_url?(url)
+      url && http?(url) && valid_domain?(url)
+    end
+
+    def http?(url)
+      url =~ /\A#{URI::regexp(['http', 'https'])}\z/
+    end
+
+    def valid_domain?(url)
+      WHITELIST.include?(Domainator.parse(url))
```
DouweM: I din't want to load Domainator just for this single check. Can you use `URI.parse(...).host` instead?

szechyjs:
@axilleas I added some docs

app/services/projects/download_service.rb
szechyjs: I'm not sure of an easy way to handle subdomains and TLDs such as `co.uk`

Perhaps do the whitelist as regex?

In the case of fogbugz the whitelist needs to allow `*.fogbugz.com`, every account has its own subdomain.

DouweM: With the below, WHITELIST can contain both string and regex entries.

```ruby
def valid_domain?
  host = URI.parse(url).host
  WHITELIST.any? { |entry| entry === host }
end
```

stanhu:
Looks like there are branch conflicts again. FYI, there is a new `ApplicationSetting` to check whether the system should allow certain importers. You'll probably need to add `fogbugz` to the list and add the appropriate checks.

szechyjs:
@stanhu I'll rebase again. The application settings stuff was added in the last rebase 2295f9a

stanhu:
Looks good to me. @DouweM, any other comments?

app/services/projects/download_service.rb
```diff
@@ -0,0 +1,43 @@
+module Projects
+  class DownloadService < BaseService
+
+    WHITELIST = [
+      /[^.]+\.fogbugz.com/
```
DouweM: Please add anchors. Right now, someone could dupe us using `whatever.fogbugz.com.evil.inc`

szechyjs:
@DouweM good catch. pushed a fix

stanhu:
So close. Another rebase needed to resolve conflicts here.

szechyjs:
rebased and squashed

stanhu:
I'm not sure if this test was previously failing on master, but master seems to be back to normal. Can you rebase again?

szechyjs:
rebased again. it seems master is failing for the same reason

stanhu:
After investigating the build failure for a while, I'm convinced it's unrelated. Filed an issue: 

https://gitlab.com/gitlab-org/gitlab-ce/issues/2469

stanhu:
Thanks, @szechyjs!

PR #9556, opened 2015-08-13
Fix changelog

pupaxxo:


pupaxxo:
I removed it bacuse it was fixed in 7.13.3

PR #9567, opened 2015-08-19
add CommitNote field expose :created_at

bigsurge:
add CommitNote field expose :created_at

### DIFF
--- a/lib/api/entities.rb
+++ b/lib/api/entities.rb
@@ -203,6 +203,7 @@ module API
       expose(:line) { |note| note.diff_new_line }
       expose(:line_type) { |note| note.diff_line_type }
       expose :author, using: Entities::UserBasic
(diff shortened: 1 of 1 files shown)


stanhu:
Thanks, @bigsurge!

Razer6:
@bigsurge Can you also add this to the API documentation??

bigsurge:
@Razer6 Yes, of course, but I do not know where to write API document, can you give me a URL？

Razer6:
The API documentation is here: https://github.com/gitlabhq/gitlabhq/tree/master/doc/api in particular for the commit notes here: https://github.com/gitlabhq/gitlabhq/blob/master/doc/api/commits.md

However, it seems the documentation already says there is a `created_at` field`. So you fixed a bug bay adding that.

PR #9591, opened 2015-08-29
Add API method for get user by ID of an SSH key

bozaro:
GL_ID environment variable in hooks can contains value like:
- user-:id
- key-:id

Unfortunately, I have not found a way to get information about the user by ID for the SSH key via current API.

### DIFF
--- a/doc/api/users.md
+++ b/doc/api/users.md
@@ -150,51 +150,6 @@ Parameters:
 }
 ```
 
-### Get SSH key with user by ID of an SSH key
-
-Get SSH key with user by ID of an SSH key. Note only administrators can lookup SSH key with user by ID of an SSH key.
-
-```
-GET /keys/:id
-```
-
-Parameters:
-
-- `id` (required) - The ID of an SSH key
-
-```json
-{
-  "id": 1,
-  "title": "Sample key 25",
-  "key": "ssh-rsa AAAAB3NzaC1yc2EAAAABJQAAAIEAiPWx6WM4lhHNedGfBpPJNPpZ7yKu+dnn1SJejgt1256k6YjzGGphH2TUxwKzxcKDKKezwkpfnxPkSMkuEspGRt/aZZ9wa++Oi7Qkr8prgHc4soW6NUlfDzpvZK2H5E7eQaSeP3SAwGmQKUFHCddNaP0L+hM7zhFNzjFvpaMgJw0=",
-  "created_at": "2015-09-03T07:24:44.627Z",
-  "user": {
-    "name": "John Smith",
-    "username": "john_smith",
-    "id": 25,
-    "state": "active",
-    "avatar_url": "http://www.gravatar.com/avatar/cfa35b8cd2ec278026357769582fa563?s=40\u0026d=identicon",
-    "web_url": "http://localhost:3000/u/john_smith",
-    "created_at": "2015-09-03T07:24:01.670Z",
-    "is_admin": false,
-    "bio": null,
-    "skype": "",
-    "linkedin": "",
-    "twitter": "",
-    "website_url": "",
-    "email": "john@example.com",
-    "theme_id": 2,
-    "color_scheme_id": 1,
-    "projects_limit": 10,
-    "current_sign_in_at": null,
-    "identities": [],
-    "can_create_group": true,
-    "can_create_project": true,
-    "two_factor_enabled": false
-  }
-}
-```
-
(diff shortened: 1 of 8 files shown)


bozaro:
Rebased on master to avoid merge conflicts

bozaro:
https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/1233

stanhu:
@bozaro Thanks, no need to send merge requests to both github.com and gitlab.com. One is sufficient.

@DouweM Is there any reason allowing lookup by key ID doesn't make sense? Someone asked about [this feature here](https://groups.google.com/forum/#!topic/gitlabhq/ht_0G6JcHOc).

bozaro:
> @DouweM Is there any reason allowing lookup by key ID doesn't make sense? Someone asked about this feature here.

In any case, I can get a list of all users consistently exhausting their IDs via current API.

The only new knowledge that I get more - see the number of keys on the user. And to know the identifiers of the keys.

I do not see this possibility of any danger.

DouweM:
- Only admin can get a user's keys via the API currently, so only admin should be able to use this lookup feature.
- I'm wondering if there's a more "Railsy"/"RESTy" way to format the URL. How about `keys/:id/user`?

bozaro:
> Only admin can get a user's keys via the API currently, so only admin should be able to use this lookup feature.

We'll let it go at that

> I'm wondering if there's a more "Railsy"/"RESTy" way to format the URL. How about keys/:id/user?

Pull request updated

stanhu:
Thanks. Can you also add a spec to `lib/api/users_spec.rb`?

bozaro:
> Thanks. Can you also add a spec to lib/api/users_spec.rb?

I'm afraid that I did not have enough skill to do so.

DouweM:
@bozaro I changed my mind a bit :) Ideally, we'd have a `keys/:id` (admin-only) that would return a JSON doc of the key, with the user embedded. That's much better than the weird `keys/:id/user` endpoint. Could you change that, and also move this into a separate `lib/api/keys.rb`? For the spec, you can copy the relevant parts from the existing `users_spec.rb`.

bozaro:
@DouweM 
I will try to make these changes.

I think it will take a couple of days. At the moment, I have no experience of programming in Ruby.
For example, I do not know how to run the spec tests.

DouweM:
@bozaro Cool. You should probably get our [dev kit](https://gitlab.com/gitlab-org/gitlab-development-kit) set up.

bozaro:
Updated. Still without specs.

bozaro:
@DouweM 
Update. Specs added.

Dev kit was extremely useful :+1:

lib/api/entities.rb
```diff
@@ -199,6 +199,11 @@ class SSHKey < Grape::Entity
       expose :id, :title, :key, :created_at
     end
 
+    class SSHKeyWithUser < Grape::Entity
```
DouweM: If you do `SSHKeyWithUser < SSHKey`, you don't need to repeat the first line of exposing :)

lib/api/keys.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    resource :keys do
+      # Get single ssh key by id. Only available to admin users.
+      #
+      # Example Request:
+      #   GET /keys/:id
+      get ":id" do
+        authenticated_as_admin!
+
+        key = Key.find_by(id: params[:id])
```
DouweM: You can simply do `Key.find(params[:id])`.

lib/api/keys.rb
```diff
@@ -0,0 +1,21 @@
...
+        not_found!('Key') unless key
```
DouweM: If you use `find` rather than `find_by`, you don't need this line.

bozaro:
Fix review notes.

doc/api/users.md
```diff
@@ -150,6 +150,51 @@ Parameters:
 }
 ```
 
+### Get SSH key with user by ID of an SSH key
```
DouweM: This should probably be moved into a separate api/keys doc since it's no longer about users.

DouweM:
@bozaro Code looks great, thanks!

doc/api/users.md
bozaro: Moved

DouweM:
@bozaro :+1:

bozaro:
Thanks :)

PR #9598, opened 2015-09-02
add image content-type header 

darkrasid:
When I use image in Markdown, I find a bug.
If I link image in MD file from gitlab raw data, it would not show on IE. 
(eg. https://gitlab.com/user_id/repo_name/raw/master/image.png)

issue : https://gitlab.com/gitlab-org/gitlab-ce/issues/2376   

It is because of content-type of headers.
If the blob is not text, the content-type is always application/octet-stream. 
And IE cannot resolve it to image so it will be cracked. (I test it IE 10, 11) 
So I add code to change content-type if blob type is image. 

```
modified:   app/controllers/projects/raw_controller.rb
```

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -36,6 +36,7 @@ v 8.0.0 (unreleased)
   - Refactored service API and added automatically service docs generator (Kirill Zaitsev) 
   - Added web_url key project hook_attrs (Kirill Zaitsev)
   - Add ability to get user information by ID of an SSH key via the API
+  - Fix bug which IE cannot show image at markdown when the image is raw file	of gitlab  
 
 v 7.14.1
   - Improve abuse reports management from admin area
--- a/app/controllers/projects/raw_controller.rb
+++ b/app/controllers/projects/raw_controller.rb
@@ -29,6 +29,8 @@ class Projects::RawController < Projects::ApplicationController
   def get_blob_type
     if @blob.text?
       'text/plain; charset=utf-8'
+    elsif @blob.image?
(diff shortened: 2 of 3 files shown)


CHANGELOG
```diff
@@ -26,6 +26,8 @@ v 8.0.0 (unreleased)
   - Make all profiles public
   - Fixed login failure when extern_uid changes (Joel Koglin)
   - Don't notify users without access to the project when they are (accidentally) mentioned in a note.
+	- Fix bug which IE cannot resolve image problem in markdown when the image file is from raw data.
```
stanhu: Fix indentation.

stanhu:
Please rebase master. There is also a new spec in `raw_controller_spec.rb`; please add an image file to that and test this change.

darkrasid:
I rebased master and add test case for the raw_controller. 

If you want to check the situation of issue please access blew link with Internet Explorer 11, 10 or lower.
(https://gitlab.com/darkrasid/testtest/blob/master/test.md)

Thanks

stanhu:
There appears to be a merge conflict; possibly with CHANGELOG? Can you rebase once more and fixed the comment message?

darkrasid:
Hi. I rebase and fix comment. 
However there is a build error. I try to find cause and find some error of last gitlab build of master. (https://github.com/gitlabhq/gitlabhq/branches) 
Please check gitlab master branch.

stanhu:
Merging this because test failure was fixed in aa75e76a0a.

PR #9601, opened 2015-09-02
Piwik url with protocol in config value

isserrano:
When having a full url with the protocol in the gitlab.yml file the javascript injected won't work.

This fix will make the config file values be like:

piwik_url: "https://piwiklocation.com/piwik"
piwik_site_id: "7"

### DIFF
--- a/app/views/layouts/_piwik.html.haml
+++ b/app/views/layouts/_piwik.html.haml
@@ -4,7 +4,7 @@
   _paq.push(["enableLinkTracking"]);
 
   (function() {
-    var u=(("https:" == document.location.protocol) ? "https" : "http") + "://#{extra_config.piwik_url}/";
+    var u="#{extra_config.piwik_url}/";
     _paq.push(["setTrackerUrl", u+"piwik.php"]);
     _paq.push(["setSiteId", "#{extra_config.piwik_site_id}"]);
     var d=document, g=d.createElement("script"), s=d.getElementsByTagName("script")[0]; g.type="text/javascript";


stanhu:
It looks like the omnibus version uses the right URL format. Do you know what happens if you configure GitLab this way? 

https://gitlab.com/gitlab-org/omnibus-gitlab/commit/1202a727aeb77667e8561747c72ec5572fe31c18

isserrano:
As it is now if you put http or https with the url in the gitlab.yml file
it fails when calling teh piwik server.

stanhu:
@Seklfreak, how did this ever work? Do you have to assume `piwik_url` doesn't contain the leading `http[s]://`? I'm surprised nobody has run into this for over a year.

Seklfreak:
@stanhu Exactly. The scheme is added in JS, because a HTTPS website can't access HTTP content. Calling it `piwik_url` was a misleading decision, I am sorry. :/

stanhu:
Perhaps to make this backwards compatible with older settings HTTP[s] should only be appended to the setting if no scheme is present.

axilleas:
That is correct. piwik gives you this js snippet to put in your html:

```js
<!-- Piwik -->
<script type="text/javascript">
  var _paq = _paq || [];
  _paq.push(['trackPageView']);
  _paq.push(['enableLinkTracking']);
  (function() {
    var u="//piwik.gitlab.com/";
    _paq.push(['setTrackerUrl', u+'piwik.php']);
    _paq.push(['setSiteId', 5]);
    var d=document, g=d.createElement('script'), s=d.getElementsByTagName('script')[0];
    g.type='text/javascript'; g.async=true; g.defer=true; g.src=u+'piwik.js'; s.parentNode.insertBefore(g,s);
  })();
</script>
<noscript><p><img src="//piwik.gitlab.com/piwik.php?idsite=5" style="border:0;" alt="" /></p></noscript>
<!-- End Piwik Code -->
```

where `piwik.gitlab.com` is the fqdn.

axilleas:
@isserrano thanks for the initiative! Closing in favor of #9813

PR #9602, opened 2015-09-02
Add API request to get slack service parameters from projects/:id/services

bpetheo:
Added a new API endpoint to retrive slack parameters so automated pull sctipts can get those without any extra config and can post to the preconfigured channel.

Related feature request: http://feedback.gitlab.com/forums/176466-general/suggestions/9590679-api-get-slack-parameters

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -31,6 +31,7 @@ v 8.0.0 (unreleased)
   - Don't notify users without access to the project when they are (accidentally) mentioned in a note.
   - Retrieving oauth token with LDAP credentials
   - Load Application settings from running database unless env var USE_DB=false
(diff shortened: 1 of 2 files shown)


stanhu:
Thanks for contributing! 6f19e8799e4c8233b01068bf6b5406b0ba1d1105 has refactored the API to support all project services, including Slack, so this PR is no longer necessary.

bpetheo:
Thanks for your response. The refactored API still lacks the ability to get these parameters, it supports only put and delete. I've extended the refactored API with a get function so these parameters can be retrived through the API from now.

The new PR can be found here: https://github.com/gitlabhq/gitlabhq/pull/9604

PR #9618, opened 2015-09-11
add repository field to issue hook data

pjknkda:
**What does this do?**
Webhooks for issue now contains repository field which is in the same manner as comment events.

**Why is this needed?**
Webhook provides the ability to update an external issue tracker, but there is no way to figure out the repository name of an issue when issue events are occured. Actually we can do it using project_id, but other webhook events provides both repository info and project_id.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -50,6 +50,7 @@ v 8.0.0 (unreleased)
   - Added service API endpoint to retrieve service parameters (Petheő Bence)
   - Add FogBugz project import (Jared Szechy)
   - Sort users autocomplete lists by user (Allister Antosik)
+  - Webhook for issue now contains repository field (Jungkook Park)
 
 v 7.14.3
   - No changes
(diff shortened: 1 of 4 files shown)


stanhu:
Thanks for the PR. Can you add a spec to `web_hook_spec.rb` for this?

pjknkda:
Do you mean `spec/models/hooks/web_hook_spec.rb`? Can you explain more detail what should I modify in that file because I'm new to Ruby and have no idea how to manage the spec file.

stanhu:
Sorry, I meant `issuable_spec.rb`. You would need to add a spec to test the `to_hook_data` call on `issue`.  See `note_data_builder_spec.rb` as an example.

stanhu:
Also, there looks like there are branch conflicts, so you may have to rebase master.

pjknkda:
I added a spec for the to_hook_data of issuable model, but am wondering this is correct way.

stanhu:
@pjknkda Can you rebase master? There seems to be some issue with your commit.

pjknkda:
Oh I missed it. Now I did rebase to master.

stanhu:
Thanks, looks good. Can you squash this PR into a single commit?

pjknkda:
Okay.

stanhu:
Thanks, @pjknkda!

PR #9632, opened 2015-09-15
Use fixed version of fogbugz gem

szechyjs:
This allows us to properly handle authentication errors.


szechyjs:
@DouweM @stanhu

stanhu:
@szechyjs Great, thanks for following up on that!

PR #9522, opened 2015-08-02
Set user as default sort order for user auto-complete lists

allistera:
Currently when Gitlab gets a list of users (For example in the Assignee field for a merge request) for a autocomplete select field it orders them by date created, it would be must more elegant if this was sorted by the users name.

It looks like this was previously the case, but the code has been changed since then:
https://github.com/gitlabhq/gitlabhq/pull/3645

PR #9634, opened 2015-09-15
Sorted autocomplete users list by name

allistera:
Currently when Gitlab gets a list of users (For example in the Assignee field for a merge request) for a autocomplete select field it orders them by date created, it would be must more elegant if this was sorted by the users name.

It looks like this was previously the case, but the code has been changed since then:
#3645

This is a re-pull request of https://github.com/gitlabhq/gitlabhq/pull/9522 - due to CI not running on it.


stanhu:
@allistera Thanks!

PR #9650, opened 2015-09-21
FogBugz Import: Closed comments may contain user data

szechyjs:


stanhu:
What happens if a closed comment is skipped? Does the import look like it's missing some information?

szechyjs:
During my initial testing closed comments appeared to have no user data, just a generated message saying the issue has been closed with a timestamp and user information.

From recent imports from FogBugz it was discovered that if a user did an equivalent of 'comment and close' the user's comment would then be a part of that auto generated comment.

By skipping the closed comments those user entered messages are lost.

stanhu:
Thanks, I misread the change.

PR #9669, opened 2015-09-26
Fix grammar in admin area labels when no labels exist

aaronsnyder:
Spoke to @dblessing in IRC re: this change. Noticed a grammar error when using GitLab, changed "...no any..." to "...not any...".

### DIFF
--- a/app/views/admin/labels/index.html.haml
+++ b/app/views/admin/labels/index.html.haml
@@ -12,5 +12,5 @@
     = paginate @labels, theme: 'gitlab'
   - else
     .light-well
-      .nothing-here-block There are no any labels yet
+      .nothing-here-block There are no labels yet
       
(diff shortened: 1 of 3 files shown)


app/views/admin/labels/index.html.haml
```diff
@@ -12,5 +12,5 @@
     = paginate @labels, theme: 'gitlab'
   - else
     .light-well
-      .nothing-here-block There are no any labels yet
-      
\ No newline at end of file
+      .nothing-here-block There are not any labels yet
```
stanhu: Perhaps it should be: "There are no labels yet"?

aaronsnyder:
Hi @stanhu, thanks for the feedback! You're right, after grepping around a bit, the standard seems to be 'there are no...' over 'there are not any'. My apologies for the oversight! 
 
I don't rebase often enough, and had trouble getting this down to a single commit. Should I nuke this branch, start over, and create a new PR?

stanhu:
@aaronsnyder It's up to you. I think you should be able to squash this into a single commit:

http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

aaronsnyder:
Hi @stanhu, you were right, I was able to squash the commits. Please let me know if there is anything else you need from me.

stanhu:
Thanks, @aaronsnyder!

PR #9671, opened 2015-09-26
Return 204 instead of 404 if no jobs waiting

janten:
See https://github.com/gitlabhq/gitlab-ci/issues/363

### DIFF
--- a/lib/api/helpers.rb
+++ b/lib/api/helpers.rb
@@ -204,6 +204,10 @@ module API
 
     # error helpers
 
+    def no_content!
+      render_api_error!('', 204)
+    end
+
     def forbidden!(reason = nil)
       message = ['403 Forbidden']
       message << " - #{reason}" if reason
--- a/lib/ci/api/builds.rb
+++ b/lib/ci/api/builds.rb
@@ -14,7 +14,7 @@ module Ci
           authenticate_runner!
           update_runner_last_contact
           required_attributes! [:token]
-          not_found! unless current_runner.active?
+          no_content! unless current_runner.active?
 
           build = Ci::RegisterBuildService.new.execute(current_runner)
 
@@ -22,7 +22,7 @@ module Ci
             update_runner_info
             present build, with: Entities::Build
           else
-            not_found!
+            no_content!
           end
         end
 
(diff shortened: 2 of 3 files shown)


janten:
Any news on this?

JulienBlancher:
+1 would be great :)

ayufan:
I expect it to be resolved in 8.3.

JulienBlancher:
Ok nice thanks !

stanhu:
Should this be closed, @ayufan?

ayufan:
@stanhu Not yet, I want to test it and possibly merge in the upcoming release.

janten:
Any news on this? 8.3 still returns the 404 error.

asziranyi:
@ayufan can you give us a status on this issue? Thanks.

cadavre:
Any info? Why return error on proper check? A must to implement.

ayufan:
@tmaczukin I think that we should be ready to merge this change. Could you verify what consequences it will have for runner older than 0.7, and also previous ruby runner? From what I think it will only lead to failure message printed in logs.

tmaczukin:
@ayufan OK, I will look on this.

tmaczukin:
@ayufan Old ruby runner was looking only for status code `201` (from some version also for `403`) while receiving new builds. Any other status code was causing a message "nothing". So for old ruby runner there is no difference between `204` and `404`.

New runner in go, before version 1.0.0, was looking for more status codes, but any other than `201` was causing only an warning/error message (and for `403` or `clientError` there is also `false` returned from the `GetBuild` method). `204` in case of versions older than 1.0.0 will be matched by `default` in the switch statement, and because of this the message will be an Error 'Checking for builds... failed'. But return of the method will be the same as for `404` code: `return nil, true`.

Since version 1.0.0 status code `204` is caught in parallel with `404` in the switch statement.

connorshea:

If you'd like to get this merged, I encourage you to open this as a Merge Request on [the GitLab.com project](https://gitlab.com/gitlab-org/gitlab-ce). Feel free to ping me (@connorshea) and I'll get the right person assigned to it.

Issue #9413, opened 2015-06-25
Commit messages not visible if emoji is present


jonathanwiesel:
I noticed something odd with the new release(s), if the commit message has an emoji, it's not rendered on the file explorer. 

![screen shot 2015-06-25 at 2 39 56 pm](https://cloud.githubusercontent.com/assets/1326781/8363374/4efcaa92-1b48-11e5-8ca8-91fd62532485.png)

NKMR6194:
Same problem is occurred on list of commits.

![2015-07-22 22 33 33](https://cloud.githubusercontent.com/assets/4254002/8826529/24016014-30c2-11e5-8d67-fe30797ae909.png)

jonathanwiesel:
Any news on this?

Razer6:
Everything after the emoji is truncated. Reproduced in 7.14 on gitlab.com /cc @tsigo

tsigo:
Related: https://gitlab.com/gitlab-org/gitlab-ce/issues/2349

Razer6:
Then I close it here. Continue discussion on GitLab.

PR #9684, opened 2015-10-01
Fix link with emoji

ysakasin:
Fix issue:
https://github.com/gitlabhq/gitlabhq/issues/9413
https://gitlab.com/gitlab-org/gitlab-ce/issues/2349

When exec `link_to_gfm(':book:Book', '/foo')`, The value of `gfm_body` is:

```
<img class="emoji" title=":book:" alt=":book:" src="path/to/emoji.png" height="20" width="20" align="absmiddle">Book
```

The result of `Nokogiri::XML::DocumentFragment.parse(gfm_body)` is:

```
<img class="emoji" title=":book:" alt=":book:" src="path/to/emoji.png" height="20" width="20" align="absmiddle">Book</img>
```

because the img tag has not an end tag and it is not XML. 
So `Nokogiri::HTML::DocumentFragment.parse(gfm_body)` is returned the expected value which is `img` tag and `Book`.


stanhu:
Thanks for looking into this. I am confused why there is `...` in the `img` tag in the first place. Also, why is there no closing `</img>` tag?

NKMR6194:
Sorry, `...` means omission. I has changed the discription.

Emoji filter return `img` tag with end slash ([this line](https://github.com/gitlabhq/gitlabhq/blob/54452412f765d9e6e6166e105db9adbc7553aec2/lib/gitlab/markdown/emoji_filter.rb#L42)) but `HTML::Pipeline` maybe remove that slash.

`</img>` tag is not usually used but I think using `</img>` is nice.

stanhu:
Hmm, sounds like `HTTML::Pipeline` should not be removing that slash?

/cc: @tsigo

NKMR6194:
I searched about `img` tag on HTML5.
I think `HTTML::Pipeline` should be removing that slash on HTML5.

* `img` tag is [void element](http://www.w3.org/TR/html5/syntax.html#void-elements). It can't have any contents and end tag.
*  `` is syntax sugar of ``. (ref: [start tags](http://www.w3.org/TR/html5/syntax.html#start-tags))

stanhu:
Oh right, that makes sense.

stanhu:
Looks like the build is failing:

```
  1) GitlabMarkdownHelper#link_to_gfm should replase commit message with emoji to link
     Failure/Error: expect(act).
     NameError:
       undefined local variable or method `act' for #<RSpec::ExampleGroups::GitlabMarkdownHelper::LinkToGfm:0x007fb8d5006488>
     # ./spec/helpers/gitlab_markdown_helper_spec.rb:112:in `block (3 levels) in <top (required)>'
```

spec/helpers/gitlab_markdown_helper_spec.rb
```diff
@@ -106,6 +106,12 @@
       act = link_to_gfm(text, '/foo')
       expect(act).to eq %Q(<a href="/foo">#{issues[0].to_reference}</a>)
     end
+
+    it 'should replase commit message with emoji to link' do
```
stanhu: Misspelling here: replase -> replace

spec/helpers/gitlab_markdown_helper_spec.rb
```diff
@@ -106,6 +106,12 @@
...
+      actual = link_to_gfm(':book:Book', '/foo')
+      expect(act).
```
stanhu: I think you meant `expect(actual)`

stanhu:
Thanks. Looks good. Can you squash this into one commit?

NKMR6194:
Thanks reviewing. I have squashed commits. Is it OK?

stanhu:
Looks good to me. @tsigo, is there any reason that parsing as HTML would cause issues? I notice some of the filters explicitly say `Returns a Nokogiri::XML::DocumentFragment`.

tsigo:
Hating myself right now for how simple this fix was :+1: 

LGTM.

stanhu:
Thanks, @NKMR6194!

NKMR6194:
:smile:

PR #9696, opened 2015-10-05
Added user preference to change layout width

gopeter:
As discussed [here](https://twitter.com/Go_Peter/status/646259409807646720), I've added an option to choose between a small and a wide application layout. Since there aren't tests for such UI helper classes/functions (like the `fluid_layout` function), I don't added tests for this feature.

To achieve this, I've added a new column to the user table which stores the layout option and extended the `fluid_layout` function which now uses the user preference.

Screenshot before:

before

Screenshot after:

after


gopeter:
I'm on the way to fix tests

gopeter:
OK, I think I need some help: the tests fail because `current_user` isn't available inside `features/steps/..` files. Changing the code to something like `current_user ? foo : bar` isn't an option because I don't want to add more code just to pass the tests. Since I don't have any experience with steps, I need some assistance to get `current_user` working inside this test files.

And I have to revert my first statement about missing tests for UI helper methods – I've just didn't found the right ones :) Sorry for confusing!

randx:
@gopeter if user is not logged in `current_user` is `nil` so calling method `current_user.layout` raise an exception. You need to add extra check if curren_user is not nil

app/helpers/page_layout_helper.rb
```diff
@@ -26,7 +26,7 @@ def sidebar(name = nil)
 
   def fluid_layout(enabled = false)
     if @fluid_layout.nil?
-      @fluid_layout = enabled
+      @fluid_layout = (current_user.layout == "wide" || enabled == true) ? true : false
```
randx: @gopeter this can be improved like 

```
@fluid_layout = (current_user && current_user.layout == "wide") || enabled
```

gopeter:
OK, thanks! 

Extra check would mean something like:
````
def fluid_layout(enabled = false)
  if @fluid_layout.nil?
    if current_user
      @fluid_layout = (current_user && current_user.layout == "wide") || enabled
    else
      @fluid_layout = enabled  
    end
  else
    @fluid_layout
  end
end
````

**Comment by randx** [comment]
@gopeter or you can look on comment above for one line solution :)

**Comment by gopeter** [comment]
Aaah, I missed `current_user` in front of the condition :)

**Comment by randx** [comment]
@gopeter you dont need to check for current user twice. It can be like this: 

```ruby
  def fluid_layout(enabled = false)
    if @fluid_layout.nil?
      @fluid_layout = @fluid_layout = (current_user && current_user.layout == "wide") || enabled
    else
      @fluid_layout
    end
  end
```

randx:
@gopeter yes :smile:

gopeter:
Tests are passing now, except one (which has nothing to do with my code I think):
````
Feature: Login form

  Scenario: I see crowd form
    ✔  Given Crowd integration enabled         # features/steps/login_form.rb:8
    ✔  When I visit sign in page               # features/steps/login_form.rb:22
    HTML screenshot: /Users/peter_goebel/Sites/Workspace/PGW/tools/gitlab-development-kit/gitlab/tmp/capybara/screenshot_2015-10-05-16-28-33.133.html
    ✘  Then I should see Crowd login form      # features/steps/login_form.rb:16

        expected to find css "#tab-crowd form" but there were no matches
        /Users/peter_goebel/.rvm/gems/ruby-2.1.6/gems/rspec-support-3.3.0/lib/rspec/support.rb:86:in `block in <module:Support>'
````

randx:
@gopeter should be random failing test. try to restart

gopeter:
Just run the test suite three times and this test is always failing.

gopeter:
****, sorry, tried to squash commits. I'm going to close this MR and create a new one.

randx:
@gopeter ok np :)

gopeter:
Tests are now passing (#9697), except four performance tests:

````
rspec ./spec/benchmarks/models/user_spec.rb:19 # User.by_login using a capitalized username should run at least 1000 iterations per second
rspec ./spec/benchmarks/models/user_spec.rb:25 # User.by_login using a lowercase username should run at least 1000 iterations per second
rspec ./spec/benchmarks/models/user_spec.rb:31 # User.by_login using a capitalized Email address should run at least 1000 iterations per second
rspec ./spec/benchmarks/models/user_spec.rb:37 # User.by_login using a lowercase Email address should run at least 1000 iterations per second
````

PR #9697, opened 2015-10-05
Added user preference to change layout width

gopeter:
As discussed [here](https://twitter.com/Go_Peter/status/646259409807646720), I've added an option to choose between a small and a wide application layout. Old MR: #9696 

To achieve this, I've added a new column to the user table which stores the layout option and extended the `fluid_layout` function which now uses the user preference.

Screenshot before:

before

Screenshot after:

after

### DIFF
--- a/app/helpers/preferences_helper.rb
+++ b/app/helpers/preferences_helper.rb
@@ -2,8 +2,8 @@
 module PreferencesHelper
   def layout_choices
     [
-        ['Small', :small],
-        ['Wide', :wide]
+        ['Fixed', :fixed],
+        ['Fluid', :fluid]
     ]
   end
 
--- a/app/helpers/preferences_helper.rb
+++ b/app/helpers/preferences_helper.rb
@@ -1,5 +1,12 @@
 # Helper methods for per-User preferences
 module PreferencesHelper
+  def layout_choices
+    [
+        ['Small', :small],
+        ['Wide', :wide]
+    ]
+  end
+
   # Maps `dashboard` values to more user-friendly option text
   DASHBOARD_CHOICES = {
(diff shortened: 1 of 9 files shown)


randx:
@gopeter looks like we forgot to exclude benchmark tests so its failing.

app/helpers/preferences_helper.rb
```diff
@@ -1,5 +1,12 @@
 # Helper methods for per-User preferences
 module PreferencesHelper
+  def layout_choices
+    [
+        ['Small', :small],
+        ['Wide', :wide]
```
randx: @gopeter how about name it `fixed` and `fluid`. I think it will better describe the setting

db/migrate/20151005150751_add_layout_option_for_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddLayoutOptionForUsers < ActiveRecord::Migration
+  def change
+    add_column :users, :layout, :integer, :default => 0
```
randx: @gopeter please use 1.9 has syntax:

```
add_column :users, :layout, :integer, default: 0
```

app/helpers/preferences_helper.rb
gopeter: For sure! :)

How can I exclude benchmark tests?

randx:
@gopeter I have a naming proposal before we merge it. What do you think?

gopeter:
Great, looks good! Fixes on the way :)

randx:
@gopeter awesome!

gopeter:
Thanks for your help! :smile:

randx:
@gopeter thanks

PR #9719, opened 2015-10-12
Small typo fix when deleting groups

teuneboon:
When I tried to delete a group I came across a huge issue that I couldn't possibly ignore. I made a group called "testgroup", and this was the message it gave me when I deleted it: `Group 'testgroup was deleted.`. That weird floating quote is obviously unacceptable for a feature used so often. The message now reads: `Group 'testgroup' was successfully deleted.`(added "successfully" cause that's also used when updating and creating groups)


stanhu:
:)  Thanks, @teuneboon!

PR #9728, opened 2015-10-14
Ignoring hidden refs when checking for repository emptiness

bozaro:
I use https://github.com/bozaro/git-as-svn project for access to git repositories via svn protocol.

This project create ref `git-as-svn/v1/master` for store:
- subversion revision to git commit mapping;
- subversion repository guid.

This ref created event on empty repository.

Before this change GitLab fail on empty project page with error:

```
ActionController::UrlGenerationError at /root/bar1

No route matches {:action=>"show", :controller=>"projects/commits", :id=>nil, :namespace_id=>#<Namespace id: 1, name: "root", path: "root", owner_id: 1, created_at: "2015-09-26 17:02:43", updated_at: "2015-09-26 17:02:43", type: nil, description: "", avatar: nil>, :project_id=>#<Project id: 11, name: "bar1", path: "bar1", description: "", created_at: "2015-09-27 05:30:51", updated_at: "2015-09-27 05:30:51", creator_id: 1, issues_enabled: true, wall_enabled: false, merge_requests_enabled: true, wiki_enabled: true, namespace_id: 1, issues_tracker: "gitlab", issues_tracker_id: nil, snippets_enabled: false, last_activity_at: "2015-09-27 05:30:51", import_url: "", visibility_level: 10, archived: false, avatar: nil, import_status: "none", repository_size: 0.0, star_count: 0, import_type: nil, import_source: nil, commit_count: 0>} missing required keys: [:id]
```

After this change it show correct page for empty repository.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -88,6 +88,7 @@ v 8.0.5
   - Correct lookup-by-email for LDAP logins
   - Fix loading spinner sometimes not being hidden on Merge Request tab switches
   - Animate the logo on hover
+  - Show "Empty Repository Page" for repository without branches (#9728)
 
 v 8.0.4
   - Fix Message-ID header to be RFC 2111-compliant to prevent e-mails being dropped (Stan Hu)
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -567,7 +567,7 @@ class Project < ActiveRecord::Base
   end
 
   def empty_repo?
-    !repository.exists? || repository.empty?
+    !repository.exists? || !repository.has_visible_content?
   end
 
   def repo
--- a/app/models/repository.rb
+++ b/app/models/repository.rb
@@ -44,6 +44,19 @@ class Repository
     raw_repository.empty?
   end
 
+  #
+  # Git repository can contains some hidden refs like:
+  #   /refs/notes/*
+  #   /refs/git-as-svn/*
+  #   /refs/pulls/*
+  # This refs by default not visible in project page and not cloned to client side.
+  #
+  # This method return true if repository contains some content visible in project page.
+  #
+  def has_visible_content?
+    !raw_repository.branches.empty?
+  end
+
   def commit(id = 'HEAD')
     return nil unless raw_repository
     commit = Gitlab::Git::Commit.find(raw_repository, id)


stanhu:
It's not clear to me why this change would resolve the error message shown above. Is the URL generator unable to create the URL because the ref name doesn't start with `refs/heads/XXX`?

bozaro:
There is two bugs in my case:

 * GitLab try to show project activity for repository without user-visible content (this change fix it);
 * GitLab can't create URL because there is not refs staring with ```refs/heads/XXX``` in this repository (this bug is not fixed).

bozaro:
I change conditional to show project activity page for repository with tags but without branches.

But activity page for tag-only repository also fail with similar error.

stanhu:
Yeah, I'm not sure if this is the right fix. A repo is empty if it has just been initialized has no commits. Whether GitLab can display any of the refs is another issue.

bozaro:
I think, show empty repository page for repository without user-visible refs is good idea.

It will almost always touch the newly created repository. I see no reasons why user can remove all branches.

In my case this hidden ref creates immediately after registering repository in git-as-svn via project_create system hooks. For empty repository it contatins only "revision 0" subversion information.

bozaro:
I tried to fix show repository without tags and failed :(

I think, show empty repository page for repository without branches is good enought: it's much better then error message.

So, I think the besh change is keep first commit 20ba3dbc855d1d4d5bbf20de1ea1b266057c69b2 commit and remove second commit c34e8fba5b11aeabdd413506880bd07883b6a6c8.

@stanhu What will be your decision?

bozaro:
@stanhu 

I see the following options:

 1. Keep as is (if somebody create create empty repository with strange refs, then he will see error page);
 1. Show "empty repository page" for repositories without branches (if somebody create tags-only repository, he will see "empty repository page");
 1. Show "empty repository page" for repositories without user visible content (branches and tags) - Looks like correct solution, but I do not have the expertise to implement it;

I think the second option is good enought.

I would like to receive any comment to make this request more appropriate. Maybe better to change the template, not a function Repository::empty.

stanhu:
@bozaro Yeah, I think a new method needs to be used instead of `Repository::empty`, such as `Repository::visible_content?`.

I also wonder whether we always want to show the `Files` page if users can create content directly with the "Create file" and "Create directory" interface now. I think it would still be useful to have instructions on how to push for an empty repo.

What do you think, @DouweM?

bozaro:
I create new method Repository::has_visible_content() and change condition in Project::empty_repo().

I tried to create a new method Project::has_visible_content(). But it looks like nonsensical:

 * This change require replace almost all usage of Project::empty_repo() to Project::has_visible_content()
 * I can't describe difference "empty repository" vs "repository without user content" for GUI;
 * Non-empty repository without user content is exotic use case and no one is going to specifically check it.

DouweM:
The addition of `visible_content?` to `empty_repo?` makes sense to me.

@bozaro Could you add a comment by `def visible_content?` to show why this is needed, i.e. why it's different from just `empty?`?

@stanhu I'm not completely sure what you mean about the Files page?

bozaro:
@DouweM 
I added comment for ```visible_content```.

DouweM:
@bozaro Cool. Can you add a changelog item as well describing the issue this fixes?

bozaro:
Added changed, rebase to master and squash

bozaro:
Thanks

PR #9729, opened 2015-10-14
ScrollTop added and button UI changed

Thirumal-Sakthivel:
Existing UI in project creation page,

new_project_page

Changed "Import project from" row buttons to icons like this,

changed_new_project_page

One more function "ScrollTop" added refer the screenshot below,

scroll_top_ui

If this code is valuable please accept this pull request.

Thanks,

### DIFF
--- a/app/assets/javascripts/pager.js.coffee
+++ b/app/assets/javascripts/pager.js.coffee
@@ -40,3 +40,13 @@
         unless @loading.is(':visible')
           @loading.show()
           Pager.getOld()
+
+$(window).scroll ->
+  if $(this).scrollTop() > 100
+    $('.scrollup').fadeIn()
+  else
+    $('.scrollup').fadeOut()
+  return
+$('.scrollup').click ->
+  $('html, body').animate { scrollTop: 0 }, 600
+  false
--- a/app/assets/stylesheets/framework/common.scss
+++ b/app/assets/stylesheets/framework/common.scss
@@ -398,3 +398,24 @@ table {
 .space-right {
   margin-right: 10px;
 }
+
+.scrollup {
+  width: 45px;
+  height: 45px;
+  position: fixed;
+  bottom: 50px;
+  right: 26px;
+  -webkit-border-radius:50%;
+  -moz-border-radius:50%;
+  border-radius: 50%;
+  display: none;
+  opacity: 0.4;
+}
+
+.scrollup.fa{
+  display: none;
+}
+
+.scrollup:hover {
+  opacity: 1;
+}
--- a/app/views/layouts/_page.html.haml
+++ b/app/views/layouts/_page.html.haml
@@ -28,3 +28,4 @@
       .content
         .clearfix
           = yield
+    %a.btn.btn-info.scrollup.fa.fa-chevron-up{:href => "#"}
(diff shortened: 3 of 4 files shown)


stanhu:
I personally find the small buttons hard to distinguish. Also, I think adding the scrollTop button makes the UI inconsistent; Either it should be done everywhere or not at all.

MauriceMohlek:
I like the idea with the small icons. These icons are well known to most of the developer.
It looks much cleaner.
You might add tool-tips to the icons.

Thirumal-Sakthivel:
@MauriceMohlek thanks for the suggestion, already added tooltips.

Thirumal-Sakthivel:
@stanhu scroll to top will be visible in all the pages.

stanhu:
Not sure how I feel about these changes--the icons seem too hard to read, and it seems excessive to me to have a special button to scroll up when I can just scroll up. :)  @randx?

Razer6:
I have a `Page-Up` key on my keyboard :smirk:

randx:
I think scrollTop button is unnecessary in most cases (you have scroll and keyboard shortcuts) while take a constant place in UI and can be annoying to have in daily usage. There are cases when it can be useful in web like long blog post with static navigation on top. But for GitLab it would be mostly negative effect.

According to import buttons I think its better to have text. Icons only looks good but it will affect user experience which is not good. Not everyone knows fogbugz/bitbucket/gitorious logo. So I rather have buttons with text. Maybe we can use smaller buttons on this page to fit in one row with text

Thirumal-Sakthivel:
Hi @randx  I agree with your point for 'ScrollTop button'.

If we use small buttons for import buttons,
smal-buttons

UI looks good but these six buttons are not constant. 

If we add one more button for import buttons,
For example, 
small-add

So, icons only the constant(for extra 7-12 buttons) solution to this issue. It is my simple suggestion.

Thanks

randx:
@Thirumal-Sakthivel maybe we should make it a dropdown instead. I dont think rendering 12 buttons makes sense

randx:
cc @skyruler

Thirumal-Sakthivel:
@randx Dropdown also affect the user experince(" if dropdown  list extends, we should scroll then see the remaining items. Not everyone knows fogbugz/bitbucket/gitorious" ) - It is my suggestion.

I created a UI( using dropdown toggle) part for this. Please refer this screenshot then share your feedback,

drop-down

Thanks.

randx:
@Thirumal-Sakthivel I think last screenshot is much better from user perspective. @skyruler what do you think?

Thirumal-Sakthivel:
@randx @skyruler If you think this is better UI, i will give you a new pull request for this. Please give your  status.

skyruler:
@Thirumal-Sakthivel I would love to create a project as fast as possible and hiding buttons with dropdown maybe not the best option in overall but it's a better solution for now. We can implement it and rework in our future releases

robodude666:
Most important order of business: why is the gitlab.com icon a heart? It should be that crazy fox thing.

Secondly, I personally prefer the row of icons; though I think the icons should be a bit bigger -- maybe the same size as the visibility level icons. It just makes the UI look a lot cleaner than having a bunch of buttons, whereas the dropdown menu looks clunky.

I DO think there should be a tooltip that appears on mouse-over with a description though, just in case a developer isn't familiar with a specific icon.

connorshea:
I don't think we're able to merge this as-is, though if you want to discuss this more you can feel free to open an issue on [the GitLab.com project](https://gitlab.com/gitlab-org/gitlab-ce). 

Thank you for the contribution!

Issue #9388, opened 2015-06-13
Broken URL in long commit message shown in dashboard


renkun-ken:
The URL, e.g. `https://gitlab.com/user/repo`, in a long commit message may be broken into `https://gitlab.co...` to be shown in the dashboard timeline. The URL does not direct to the right location. The correct design is, the URL text should be displayed as `https://gitlab.co...` but its `href` should be the whole URL in the commit message.

fpgentil:
Which version of Gitlab are you running? I wasn't able to reproduce here. 

![bildschirmfoto 2015-06-14 um 22 10 35](https://cloud.githubusercontent.com/assets/356881/8151375/58ff6322-12e2-11e5-93fc-5f2177cf531f.png)
![bildschirmfoto 2015-06-14 um 22 11 18](https://cloud.githubusercontent.com/assets/356881/8151376/59028e44-12e2-11e5-9229-0211ccc7e1a1.png)

renkun-ken:
I'm using gitlab.com and I mean the commit message in the dashboard (or homepage). It's like

1a2b3c4d some commit message here but cannot be fully shown [https://gitlab.co](https://gitlab.co)...

jvanbaarsen:
We are currently in the process of closing down the issue tracker on GitHub, to prevent duplication with the GitLab.com issue tracker. Since this is an older issue I'll be closing this for now. If you think this is still an issue I encourage you to open it on the [GitLab.com issue tracker](https://gitlab.com/gitlab-org/gitlab-ce/issues).

PR #9742, opened 2015-10-16
Broken URL in long commit message

gentilfp:
As mentioned here #9388, when a commit contains a long commit message including a URL, after truncating to be shown in dashboard, the URL is broken, therefore, causing a broken link.

**Commit example**
bildschirmfoto 2015-10-16 um 17 17 59

**Broken URL**
bildschirmfoto 2015-10-16 um 17 19 42
bildschirmfoto 2015-10-16 um 17 19 34

**Fixed URL**
bildschirmfoto 2015-10-16 um 17 20 30
## 

My first idea was extract the full URLs in the `events_helper.rb` and send them (if more then one) to `gfm`. I'm still now sure if this is a good approach. I'd like some **feedbacks** before writing some tests, do you think it's worth to keep the URL or simply remove the `<a href=""></a>` if the URL is not valid?

Thanks in advance :smile:

### DIFF
--- a/app/helpers/events_helper.rb
+++ b/app/helpers/events_helper.rb
@@ -181,6 +181,10 @@ module EventsHelper
     sanitize(text, tags: %w(a img b pre code p span))
   end
 
+  def event_commit_urls(message)
+    URI.extract(message)
+  end
+
   def event_commit_title(message)
     escape_once(truncate(message.split("\n").first, length: 70))
   rescue
--- a/app/views/events/_commit.html.haml
+++ b/app/views/events/_commit.html.haml
@@ -2,4 +2,4 @@
   .commit-row-title
     = link_to truncate_sha(commit[:id]), namespace_project_commit_path(project.namespace, project, commit[:id]), class: "commit_short_id", alt: ''
     &middot;
-    = gfm event_commit_title(commit[:message]), project: project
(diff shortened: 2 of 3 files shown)


stanhu:
I see two options that might work:

1. Truncate the whole link and replace it with `...`
2. Truncate the link text (e.g. `https://git...`) but keep the URL correct

This initial approach seems reasonable to me if it can do the job.

/cc: @tsigo

tsigo:
I think I like option 2 the best actually, if the URL can remain functional.

fpgentil:
I also think I like option 2 because if I have a long url for example:
```
1a2b3c4d | fix https://www.google.com/search?site=&tbm=isch&source=hp&biw=1680&bih=890&q=brazil&oq=brazil&gs_l=img.3..0l10.4271.4863.0.4952.6.6.0.0.0.0.261.719.2j2j1.5.0....0...1ac.1.64.img..1.5.717.9z4bbTpsGug#imgrc=JZvVXBvEiBxiHM%3A
```

it would be all replaced to `...`, like:
```
1a2b3c4d | fix ...
```

it just looks weird :smile:  I'll work more on this code, write some tests and get back here, thanks a lot so far @tsigo and @stanhu

jvanbaarsen:

PR #9743, opened 2015-10-17
Exceptions

deanpcmad:
Hey, I've wanted for feature for a while and I finally had a bit of time to work on it.

This feature adds Exception logging, from the Airbrake client, into GitLab. The code is fairly basic at the moment as I have copied most of it from an app I built which has been tested.

It still needs a bit of work done and I need to just write some tests for it but I wanted to see what people thought about it first. Thanks!


stanhu:
I like the general idea, although I think GitLab.com uses New Relic to log exceptions. It would be ideal if an admin could choose what exception handler to use: New Relic, Sentry, Airbrake, etc. How would this PR address that?

deanperry:
This PR is to add exception logging into Gitlab, not for exceptions to be sent to other loggers. Here are some screenshots to help explain it a bit better

screen-shot-2015-10-17-15-27-13-30aknlymsw
screen shot 2015-10-17 at 15 27 37
screen shot 2015-10-17 at 15 27 57
screen shot 2015-10-17 at 15 28 05

stanhu:
Oh, that's interesting--not what I was expecting. Looks promising. What would happen if an exception occurred and an admin wanted to see it in New Relic? Would this PR intercept that exception before New Relic and thus prevent New Relic from sending it out? Or would they work independently?

deanperry:
That really depends on how that app is setup. This only supports the Airbrake client at the moment as they have a lot of [clients/trackers](https://airbrake.io/languages). I don't have any experience with New Relic exceptions but I'd imagine they would both log exceptions to each service

stanhu:
@deanperry I would make sure that New Relic still can co-exist here, as third-party exception trackers may actually be more effective at tracking errors. I worry that the GitLab database could be overwhelmed by large number of exceptions.

@randx, what do you think of this feature?

deanperry:
Yes got should be able to, again this is to allow exceptions to be sent
directly to Gitlab instead of an external service like Airbrake.

E.g all my project details are stored in gitlab. I want exceptions from my
app to go to Gitlab so I have a central place for all my project. I can
also, once I add it, easily create issues/tickets from exceptions

jvanbaarsen:

PR #9751, opened 2015-10-19
Add ability to create directories in the editor

binford2k:
Simply type a name with a `/` directory separator and new directories
will be created. This does not do the fancy UI work that github.com
does, but it will get the job done.

I could not find tests for file creation, so I didn't add a test for
this slight behaviour modification. I did test directory traversals
though, using both absolute paths like `/tmp/foo.txt` and relative paths
like `../../foo.txt`. Neither case escaped the repository, though
attempting to traverse with a relative path resulted in a 500 error that
did not affect application stability upon reload.

### DIFF
--- a/app/services/files/create_service.rb
+++ b/app/services/files/create_service.rb
@@ -16,10 +16,10 @@ module Files
         )
       end
 
-      unless @file_path =~ Gitlab::Regex.file_name_regex
+      unless @file_path =~ Gitlab::Regex.file_path_regex
         raise_error(
           'Your changes could not be committed, because the file name ' +
-          Gitlab::Regex.file_name_regex_message
+          Gitlab::Regex.file_path_regex_message
         )
       end
 
--- a/app/services/files/create_service.rb
+++ b/app/services/files/create_service.rb
@@ -9,6 +9,13 @@ module Files
     def validate
       super
 
+      if @file_path =~ Gitlab::Regex.directory_traversal_regex
+        raise_error(
+          'Your changes could not be committed, because the file name ' +
+          Gitlab::Regex.directory_traversal_regex_message
+        )
+      end
+
       unless @file_path =~ Gitlab::Regex.file_name_regex
         raise_error(
           'Your changes could not be committed, because the file name ' +
--- a/app/services/files/create_service.rb
+++ b/app/services/files/create_service.rb
@@ -9,9 +9,7 @@ module Files
     def validate
       super
 
-      file_name = File.basename(@file_path)
-
-      unless file_name =~ Gitlab::Regex.file_name_regex
+      unless @file_path =~ Gitlab::Regex.file_name_regex
         raise_error(
           'Your changes could not be committed, because the file name ' +
           Gitlab::Regex.file_name_regex_message
--- a/lib/gitlab/regex.rb
+++ b/lib/gitlab/regex.rb
@@ -52,7 +52,7 @@ module Gitlab
     end
 
     def file_path_regex
-      @file_name_regex ||= /\A[a-zA-Z0-9_\-\.\/]*\z/.freeze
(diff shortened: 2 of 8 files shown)


stanhu:
Thanks for adding this. Couple of items:

1) Yes, the recent support for adding directories alone should resolve the relative paths fine. There is an error if the user enters in `../foo.txt` from the root directory; you may want to catch that.

2) The UI probably needs to be expanded to add a longer filename:

image

3) There are Spinach specs in `gitlab/features/project/source/browse_files.feature` for adding files. You should add some there, as well as perhaps a new spec for `create_service.rb`.

4) I always wondered what the issue was with restricting the filename characters. This prevents Unicode filenames from being created. Should we loosen that restriction?

binford2k:
> 4) I always wondered what the issue was with restricting the filename characters. This prevents Unicode filenames from being created. Should we loosen that restriction?

Totally agree. But that doesn't belong in this PR.

lib/gitlab/regex.rb
```diff
@@ -44,13 +44,20 @@ def project_path_regex_message
...
+      @file_name_regex ||= /\A[a-zA-Z0-9_\-\.\/]*\z/.freeze
     end
 
     def file_name_regex_message
-      "can contain only letters, digits, '_', '-' and '.'. "
+      "can contain only letters, digits, '_', '-' and '.'. Separate directories with a '/'. "
     end
 
+    def directory_traversal_regex
+      @directory_traversal_regex ||= /\.{2}/.freeze
```
stanhu: FYI, gitlab_git will resolve `foo/../bar` properly to `/bar`. It might still make sense to keep this check, though.

stanhu:
Looks good to me. @DouweM, any comments?

app/services/files/create_service.rb
```diff
@@ -9,9 +9,14 @@ def commit
     def validate
       super
 
-      file_name = File.basename(@file_path)
+      if @file_path =~ Gitlab::Regex.directory_traversal_regex
```
DouweM: @stanhu I don't think we need this check if gitlab_git already does the right thing to keep us inside the repository.

lib/gitlab/regex.rb
```diff
@@ -44,13 +44,20 @@ def project_path_regex_message
 
 
     def file_name_regex
-      @file_name_regex ||= /\A[a-zA-Z0-9_\-\.]*\z/.freeze
+      @file_name_regex ||= /\A[a-zA-Z0-9_\-\.\/]*\z/.freeze
```
DouweM: This is used to validate the snippet file name as well, which shouldn't contain slashes. Please create a new `file_path_regex` and use it here and also to validate directory names in the CreateDirService.

app/services/files/create_service.rb
binford2k: It already prevents escaping the repo, but it does cause an HTTP 500 instead of gracefully handling the exception. This prevents that.

DouweM: Can't we rescue from the `Repository::InvalidBlobName` error that Gitlab::Git raises, then?

binford2k: I suppose we could, but that seems like a larger refactor. You wouldn't want to catch some invalid names with the file regex and then some others by rescuing an exception.

DouweM: @binford2k Fair enough!

DouweM:
@binford2k Can you add a changelog item?

binford2k:
Done. Wasn't sure what format you'd like. Shout at me if you'd like it changed. Let me know if you want these squashed.

DouweM:
I squashed and added your name to the changelog entry: 0ea38dc519b86d2bd2e14f1df1baf0fffc043af6.

PR #9754, opened 2015-10-20
Doc: cronjob should be within user git

artem-sidorenko:


stanhu:
Looks good to me. @jacobvosmaer, what do you think?

jacobvosmaer:
I think this whole section should be removed, we do not use `whenever` anymore.

artem-sidorenko:
@jacobvosmaer what is the new way to have the CI schedules triggered? I couldn't find anything about this in migration/installation/update docs. As far as I understand https://gitlab.com/gitlab-org/gitlab-ce/issues/2989 it still used

Do I miss something?

jacobvosmaer:
@ayufan can you answer the question above?

stanhu:
Looks like it was removed in 468c355910195. See https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/1538.

stanhu:
Superseded by ab72c12028072.

artem-sidorenko:
@stanhu thx!

PR #9757, opened 2015-10-22
Add ignore white space option in merge request diff

yms9654:
fix this issue(https://gitlab.com/gitlab-org/gitlab-ce/issues/1393).
It supports a ?w=1 option on the URL when viewing diffs.

When I call reload_content method on MergreRequsetDiff, it change state to :collected.
But collected? method return false..
So I add a collected? method on MergeRequestDiff class.


stanhu:
Doesn't this have the side effect of making the merge request look like whitespace wasn't changed for all subsequent users?

yms9654:
Thank you for your feedback!

That's why I put this code on app/controllers/projects/merge_requests_controller.rb:69
@merge_request_diff.reload_content

I tested this code and if there are no 'w' option, it will show diff with white spaces.

stanhu:
What I mean is that the state of `merge_request_diff` in the database will be changed depending on the last user. This isn't the right behavior because whitespace *is*, in fact, part of the "true" diff. This would cause issues, for example, with users accessing merge request content via the API. Since the API accesses the content via the models and bypasses the controllers, the diff would change depending on the last user.

One way we could to make this work is add a `merge_request_diff_no_whitespace` field that is lazily loaded when the `w=1` parameter is specified.

yms9654:
Oh, I got your point.
I think your idea is a good way to fix this.
I'm working on it! :smiley:

stanhu:
Thanks for working on this, @yms9654! This seems to be a [much requested feature](http://feedback.gitlab.com/forums/176466-general/suggestions/4230569-tune-diff-output-in-merge-request).

stanhu:
Also, an easier method might just be to post-process the diff output and strip lines that just have whitespace changes.

randx:
I think the correct way to do is to ignore whitespaces on moment when we render diff. It has nothing to do with how data is stored. Basically its matter of how we render html

yms9654:
Hi~ @stanhu @randx.
Please check my new commit.

I added a  diffs_no_whitespace method on MergeRequestDiff class.
Then call this method when _diffs.html.haml view rendering.
I already check API call, it is working properly now.

stanhu:
@yms9654 Thanks for getting this to work. Is it possible to do what @randx suggested and use the existing `diffs` and filter out whitespace from the diff itself?

yms9654:
@stanhu It is possible, but I think it is more complicated.

1. I have to strip lines on diff.
2. check same line to remove.
3. check this lines is not relevant with real diffs(because I have to show relevant lines).
4. or I remove relevant lines(if it relevant with white space diff).
5. update addition and deletion numbers.
6. something I missed..

I always thank for your helping! :smiley:

randx:
@stanhu @yms9654 I still think its better to go with rendering than storing same diff twice with and without white spaces. Reasons are: 

* waste of storage
* every time diff changes you need to rebuild it twice. It is a performance question since merge request diffs get rebuild quite often for active respositories. 
* it feels wrong. Just imagine a case when we want be able to ignore newlines in diff. Following this logic we should store: diff, diff w/o whitespace, diff w/o newlines. And if we want to see diff without whitespaces and newlines? Then we should add extra field for diff w/o whitespace and w/o newline.

silversens:
@yms9654 Thanks a lot for working on this! I can't tell you how much my all team has been waiting for this feature to exist...

SvenBrunk:
@randx so... if this feels wrong, what is the original reason to store the diff in the database and why is that reason not valid for other kinds of diff?

randx:
@SvenBrunk the point is to store only one kind of diff (the one that git provides by default) and then makes all filters in rendering logic. 

Imagine you save a movie to your PC and it saves several versions of it: 
* movie - 10GB
* same movie with french subtitles - 10GB
* same movie with german subtitles 10GB

In total you get a movie with 30GB size even if you don't need any subtitles. 

Much better is to do it this way: 

*  original movie - 10GB
*  french subtitles - few KB
*  german subtitles - few KB

And apply subtitles to movie while rendering.

randx:
@yms9654 am I right that in latest version of this PR you removed storing it in database and just get it from diff?

yms9654:
@randx Yes you are right. I do not store in database.
I just add a instance variable to store that.
But I think if diff data was changed, it could make some bug.
So I can remove this variable. What do you think about that?

randx:
@yms9654 variable exists only during request. So I think it acceptable.  I think we can accept your merge request if you make few changes: 

* Move CHANGELOG item to 8.2
* Document `w=1` feature somewhere. Otherwise it will be rarely known/used only among people who need it
* Add w=1 support for commit  and compare pages too. As user I would expect it to work in everywhere

Is it something you would like to do?

yms9654:
@randx OK! It is a good news for me, thanks.
I'm working on it.

randx:
@yms9654 awesome. ping me when you need review

yms9654:
@randx 'Document w=1 feature somewhere' <- that means add some message on the html page?

randx:
@yms9654 I think docs will be better place until we make a proper `tips` system http://doc.gitlab.com/ce/workflow/merge_requests.html

yms9654:
@randx Please review my new commit. :smiley:

stanhu:
One issue is that the diff is computed on-the-fly, so for a large change the server may just timeout. That is why parsing the existing `merge_requst_diff` may be desirable.

yms9654:
@stanhu I call diff_no_whitespace method only if @merge_request_diff state is :collected.
It means that diff size is not over DIFF LIMIT.
So I think It could improve next time.

app/models/merge_request_diff.rb
```diff
@@ -47,6 +47,23 @@ def diffs
     @diffs ||= (load_diffs(st_diffs) || [])
   end
 
+  def diffs_no_whitespace
+    # Update ref for merge request
+    merge_request.fetch_ref
```
stanhu: Is this necessary? This should be updated by `RefreshService.` I would expect that `diffs_no_whitespace` should not cause any writes to the system.

yms9654: I think it is necessary.
`RefreshService` is refreshing all MR related with new pushed branch, even from forked projects.
But I want to update only this view's diff.

randx: @yms9654 I think what @stanhu is trying to say is that there is no need to do `fetch_ref ` since it will be done by GitLab anyway on new push. So this action is redundant

yms9654:
@randx I see now, thank you for explain. :smiley:
@stanhu I had a misunderstanding, sorry about that.
I update my code. remove unnecessary call fetch_ref.

yms9654:
@randx could you review new commit?

app/controllers/projects/compare_controller.rb
```diff
@@ -13,8 +13,11 @@ def show
     base_ref = Addressable::URI.unescape(params[:from])
     @ref = head_ref = Addressable::URI.unescape(params[:to])
 
+    if params[:w] == '1'
+      diff_options = { ignore_whitespace_change: true }
```
stanhu: Let's move this to one line:

```ruby
diff_options = { ignore_whitespace_change: true } if params[:w] == '1'
```

app/services/compare_service.rb
```diff
@@ -3,7 +3,7 @@
 # Compare 2 branches for one repo or between repositories
 # and return Gitlab::CompareResult object that responds to commits and diffs
 class CompareService
-  def execute(source_project, source_branch, target_project, target_branch)
+  def execute(source_project, source_branch, target_project, target_branch, diff_options={})
```
stanhu: I think the style has been to add spaces here:

```ruby
diff_options = {}
```

spec/controllers/projects/compare_controller_spec.rb
```diff
@@ -23,6 +23,19 @@
...
 
+  it 'compare should show some diffs with ignore whitespace change option' do
+    get(:show,
+        namespace_id: project.namespace.to_param,
+        project_id: project.to_param,
+        from: ref_from,
+        to: ref_to,
+        w: 1)
+
+    expect(response).to be_success
```
stanhu: Is there a way to verify that commits with whitespace changes aren't actually returned?

yms9654: @stanhu could you tell me how to add a new branch(just whitespace diff) in compare_controller_spec.rb?

randx:
@stanhu thank you for review. Would you mind if I leave this PR  to you?

yms9654:
@stanhu I found this https://gitlab.com/gitlab-org/gitlab-test
I want to add a new branch(just whitespace change) but I don't have a permission.
Could you help this? or another way to verify that test case?

stanhu:
@yms9654 You can fork that repository, add your changes, and submit a MR.

yms9654:
@stanhu OK, I submitted a MR. Pleas check it.
MR diff test is now available, but to test compare commit  need that MR be accepted.

stanhu:
@randx Happy to take this one.

@yms9654 Done!

yms9654:
@stanhu I submit new MR to gitlab-test. sorry for my mistake.

silversens:
Thanks again @yms9654.

Do you think it would be a good idea to make this option an user setting ?

It could easily become boring to manually always set the w=1.

yms9654:
@silversens It's my pleasure!
And you are right, it is not a good UX.
I think It could be on web page option, if there were many people want that.

yms9654:
@stanhu Please review new commit.
I update conventions you told, and compare_controller_spec.rb to test actual diff contents.

stanhu:
@yms9654 Looks great. Can you rebase master, since it seems that there are conflicts? (Most likely in CHANGELOG).

yms9654:
@stanhu I did rebase. Thanks a lot for many helps! :smile:

stanhu:
Thanks, @yms9654!

silversens:
@yms9654 I just tried out, thanks a lot! Although it is not usable on the commit page (.../commit/9bef4d473a8409f1fb5103e37775612af40dffb7?w=1). It would be great because this page compares with previous commit.

yms9654:
@silversens I see that now. It is a good idea and it looks necessary.
@stanhu How do you think about that? If you are agree, I will make this with new MR.

stanhu:
@yms9654 Yes, I think it would be good to have that. It would also be nice to have a checkbox that allows the user to select between ignoring whitespace, but that can be left to another PR. :)

silversens:
@stanhu @yms9654 great! If I could push a little more far: the last but not least nicest thing about this feature would be to have an option in user profile to choose to have this behavior by default or not.
But that also could be another PR ^^

doc/workflow/merge_requests.md
```diff
@@ -38,3 +38,15 @@ To check out a particular merge request:
 ```
 $ git checkout origin/merge-requests/1
 ```
+
+## Ignore whitespace changes in Merge Request diff view
+
+![MR diff](merge_requests/merge_request_diff.png)
+
+It you add `w=1` option to URL, you can see diff without whitespace changes.
```
jaisharma639: Does this mean adding a request parameter of w=1 like below?
https://github.com/repo1/server/compare/master...jaisharma639:master-diff?w=1
I did this but it doesn't work @yms9654

PR #9758, opened 2015-10-22
Rename API::APIHelpers -> API::Helpers;

huacnlee:
Rails Autoload find file to require is use , APIHelpers -> api_helpers.rb, not helpers.rb;

PR #9762, opened 2015-10-23
Fix API::APIHelpers -> API::Helpers again for master

huacnlee:
Rails Autoload find file to require is use , APIHelpers -> api_helpers.rb, not helpers.rb;


huacnlee:
Ref: #9758

PR #9770, opened 2015-10-25
List generic commit statuses individually in MergeRequests#show

dsander:
The GitLab 8.1 added the commit status API, this is my attempt to expose the external commit statuses to the user.

This changes the status indicator of a MergeRequest, before all CommitStatuses were grouped into one status which did not show information about the status of external (GenericCommitStatuses) statuses. Now CI::Build statuses are still grouped together, but GenericCommitStatuses are shown individually including the description and link to the status/build information.

**Before**, one could not see which commit status failed and which succeeded:
screenshot 2015-10-25 11 12 16

**After**, external commit statuses are shown individually:
out
screenshot 2015-11-14 17 30 44
screenshot 2015-11-14 17 30 51
screenshot 2015-11-14 17 31 06
screenshot 2015-11-14 18 22 28
screenshot 2015-11-14 18 22 35

### DIFF
--- a/app/helpers/ci_status_helper.rb
+++ b/app/helpers/ci_status_helper.rb
@@ -54,4 +54,8 @@ module CiStatusHelper
     project.runners.blank? &&
       Ci::Runner.shared.blank?
   end
+
+  def ci_widget_summary(commit)
+    "#{commit.statuses_with_matching_status} of #{commit.latest_generic_statuses.length + 1} checks #{commit.human_status}"
+  end
 end
--- a/app/models/ci/commit.rb
+++ b/app/models/ci/commit.rb
@@ -121,8 +121,25 @@ module Ci
       @latest_statuses ||= statuses.latest.to_a
     end
 
+    def latest_generic_statuses
+      @latest_generic_statuses ||= latest_statuses.select { |s| s.is_a?(GenericCommitStatus) }
+    end
+
+    def latest_ci_statuses
+      @latest_ci_statuses ||= latest_statuses.select { |s| s.is_a?(Ci::Build) }
+    end
+
     def latest_builds
       @latest_builds ||= builds.latest.to_a
+    end
+
+    def statuses_with_matching_status
+      matching = if running? || pending?
+                   ['running', 'pending']
+                 else
+                   [status]
+                 end
+      (matching.include?(ci_status) ? 1 : 0) + latest_generic_statuses.count { |s| matching.include?(s.status) }
     end
 
     def latest_builds_for_ref(ref)
@@ -134,11 +151,15 @@ module Ci
     end
 
     def status
-      if yaml_errors.present?
-        return 'failed'
-      end
+      @status ||= status_for(latest_statuses)
+    end
 
-      @status ||= Ci::Status.get_status(latest_statuses)
+    def human_status
+      success? ? 'succeded' : status
+    end
+
+    def ci_status
+      @ci_status ||= status_for(latest_ci_statuses)
     end
 
     def pending?
@@ -224,6 +245,14 @@ module Ci
       return if self.yaml_errors?
       self.yaml_errors = error
       save
+    end
+
+    def status_for(scope)
+      if yaml_errors.present?
+        return 'failed'
+      end
+
+      Ci::Status.get_status(scope)
     end
   end
 end
--- a/app/assets/javascripts/merge_request_widget.js.coffee
+++ b/app/assets/javascripts/merge_request_widget.js.coffee
@@ -9,6 +9,7 @@ class @MergeRequestWidget
   #
   constructor: (@opts) ->
     modal = $('#modal_merge_info').modal(show: false)
+    $('.mr-state-widget .mr-widget-heading a').click @toggleCiDetails
 
   mergeInProgress: (deleteSourceBranch = false)->
     $.ajax
@@ -58,3 +59,12 @@ class @MergeRequestWidget
 
   setMergeButtonClass: (css_class) ->
     $('.accept_merge_request').removeClass("btn-create").addClass(css_class)
+
+  toggleCiDetails: (e) ->
+    e.preventDefault()
+    if $('.mr-widget-ci-details').is(':visible')
+      $('.mr-widget-ci-details').slideUp(250)
+      $(e.currentTarget).text('Show details')
+    else
+      $('.mr-widget-ci-details').slideDown(250)
+      $(e.currentTarget).text('Hide details')
--- a/app/assets/stylesheets/pages/merge_requests.scss
+++ b/app/assets/stylesheets/pages/merge_requests.scss
@@ -119,6 +119,15 @@
   .ci-coverage {
     float: right;
   }
+
+  .mr-widget-ci-details {
+    display: none;
+    background: white;
+
+    .ci_widget {
+      padding: 10px 15px;
+    }
+  }
 }
 
 .mr_source_commit,
--- a/app/views/projects/merge_requests/widget/_heading.html.haml
+++ b/app/views/projects/merge_requests/widget/_heading.html.haml
@@ -1,15 +1,27 @@
 - if @ci_commit
   .mr-widget-heading
     .ci_widget{class: "ci-#{@ci_commit.status}"}
-      = ci_status_icon(@ci_commit)
-      %span
-        Build
-        = ci_status_label(@ci_commit)
-      for
+      = ci_icon_for_status(@ci_commit.status)
+      = ci_widget_summary(@ci_commit)
+      = link_to "Show defails", '#'
+  .mr-widget-ci-details
+    .ci_widget{class: "ci-#{@ci_commit.ci_status}"}
+      = ci_icon_for_status(@ci_commit.ci_status)
+      %span CI Build #{@ci_commit.ci_status}
       = succeed "." do
         = link_to @ci_commit.short_sha, namespace_project_commit_path(@merge_request.source_project.namespace, @merge_request.source_project, @ci_commit.sha), class: "monospace"
       %span.ci-coverage
-      = link_to "View details", builds_namespace_project_merge_request_path(@project.namespace, @project, @merge_request), class: "js-show-tab", data: {action: 'builds'}
+      = link_to "View build details", builds_namespace_project_merge_request_path(@project.namespace, @project, @merge_request), class: "js-show-tab", data: {action: 'builds'}
+    - @ci_commit.latest_generic_statuses.each do |status|
+      .ci_widget{class: "ci-#{status.status}"}
+        = ci_icon_for_status(status.status)
+        %span
+          - if status.description.present?
+            #{status.name} #{status.status}: #{status.description}.
+          - else
+            #{status.name} #{status.status}.
+          - if status.target_url.present?
+            = link_to "View details", status.target_url, target: '_blank'
 
 - elsif @merge_request.has_ci?
   - # Compatibility with old CI integrations (ex jenkins) when you request status from CI server via AJAX


stanhu:
Thanks for your contribution! For builds with many steps (e.g. GitLab itself), this change seems like it could fill the page with commit status messages. @randx, what do you think?

dsander:
Hi @stanhu,

actually the "internal" GitLabCI build steps are not show individually, just the ones reported from external services (GenericCommitStatus). 

I did the change because we use a tool like coveralls which tracks and reports code coverage changes, previously it just added comments to the merge request, but being able to use the commit statuses is a lot less intrusive (and less annoying).

randx:
@stanhu @dsander if it does not change behaviour of "internal" GitLabCI builds - then I think it makes sense to show each status. @ayufan what do you think?

ayufan:
@randx I'm not sure. It would make sense to show the failed statuses. I wouldn't show only external statuses and hide the builds.

Maybe it would make sense to have the Tab with Builds as we have when we view commit page?

dsander:
@ayufan My idea was to show the aggregated status for the build of each service to give a quick overview of MR status (treating GitLab CI as a service even though it is now integrated). The services can then expose detailed information about the build (separate jobs, coverage, code quality) in their own UI via the detail link.

jvanbaarsen:
I think it makes sense to show the status of external services as well. For example, I want to see when HoundCI says the build failed (Although I know hound doesn;t support GL just yet). Also, even you for example use GL CI and Semaphore, it would make sense to show both the statuses here.

randx:
@ayufan I leave this PR to you. Decide what is best and merge when ready

Razer6:
From UI point of view, it can get really messy if there are multiple different checks. Maybe group them and expand them on click?

dsander:
@Razer6 Sounds like a good idea to me, would you only group them when there is more then one check or always have a summary header?

Razer6:
I think making it always axpandable makes it more consistent. @randx @ayufan What's your opinion?

ayufan:
@Razer6 I'll look into it when we finish current release cycle.

randx:
@Razer6 yes I would like to see it grouped with expand button. 

>  always axpandable makes it more consistent

+1

dsander:
I updated the PR description with new screenshots, what do you think?

stanhu:
Looks great to me! Can you rebase master to remove the conflicts?

Razer6:
Great work :+1:

randx:
:+1:

app/helpers/ci_status_helper.rb
```diff
@@ -51,4 +51,8 @@ def render_ci_status(ci_commit)
       ci_status_icon(ci_commit)
     end
   end
+
+  def ci_widget_summary(commit)
+    "#{commit.statuses_with_matching_status} of #{commit.latest_generic_statuses.length + 1} checks #{commit.human_status}"
```
stanhu: Why is the +1 here needed? Does this assume `ci_status` is in the list?

dsander: The `latest_ci_statuses` are always grouped into one 'check' and the whole block is show shown if `@merge_request.ci_commit` is true, I assumed this means that there is at least one CI build.

app/models/ci/commit.rb
```diff
@@ -125,10 +125,27 @@ def latest_statuses
       @latest_statuses ||= statuses.latest.to_a
     end
 
+    def latest_generic_statuses
```
ayufan: We should introduce method for CommitStatus to indicate whether the it should be shown on MR page.

app/helpers/ci_status_helper.rb
ayufan: Why distinguish generic statuses from Ci::Builds?

app/models/ci/commit.rb
```diff
@@ -125,10 +125,27 @@ def latest_statuses
...
       @latest_builds ||= builds.latest.to_a
     end
 
+    def statuses_with_matching_status
+      matching = if running? || pending?
+                   ['running', 'pending']
+                 else
+                   [status]
+                 end
+      (matching.include?(ci_status) ? 1 : 0) + latest_generic_statuses.count { |s| matching.include?(s.status) }
```
ayufan: What exactly it does?

ayufan:
@dsander I'm not sure about distinguishing generic_statuses from builds.

app/helpers/ci_status_helper.rb
ayufan: Also this method is crazy - no idea what is happening here ;)

dsander: > Why distinguish generic statuses from Ci::Builds? The builds are also "checks", but for different test suites.

I'll use the TravisCI terminology of builds which contain many jobs. I do not think that we should list the status of every job but rather the build status. Showing details of the build (the individual jobs) should be the responsibility of the service (GitlabCI, TravisCI, Coveralls, etc) that reported the status.

> Also this method is crazy - no idea what is happening here ;)

Don't know what is so crazy about it, but of course are open for simplification suggestions. 
`commit.statuses_with_matching_status` returns the amount of statuses that match the current status of the MR itself. 
`commit.latest_generic_statuses.length` is the amount of generic statuses, `+ 1` because no matter how many CI builds there are they are always grouped into one status. 
`commit.human_status` changes `success` to `succeeded` because "4 of 4 checks success" would not be a real sentence.

app/models/ci/commit.rb
dsander: See above for a general explanation, `running` and `pending` need to be handled that way because `Ci::Status.get_status` returns `running` if any of the statuses are either running or pending

dsander:
@ayufan I thought that was already agreed upon? If that is not the case I will wait until you have decided it internally :smile:

dsander:
Any update on this? I noticed you added the build summary for merge requests but did not find the `description` of external statuses anywhere.

ayufan:
@dsander 

This was added by @DouweM, but I'm not sure about our plan for showing description. I don't see where we can present this.

@DouweM I think that this MR is stuck now, do you have any thoughts what we should due with this one from the UX perspective after adding the Builds tab?

@dsander Thanks for your contribution. We should discuss what we can do to show the description. I feel that it would be nice to present it somewhere.

DouweM:
@ayufan I think we should show the description for every build, but at that point it is no longer feasible to present the builds as simple rows in a table. We would need more specialized list items for builds, that can have two lines of content and make more effective use of the available space. This would be something for @creamzy to work on.

dsander:
@ayufan Thanks for the update!

What do you think about not showing the `GenericCommitStatus` in the builds tab? They can be anything, coverage reports, code quality or the build status of an external test runner. This would also make the build tab consistent with the projects builds page.
Without the GenericCommitStatus in the list, I think the accordion + link to the builds tab for the CI status would be a valid solution from the UX perspective.

app/models/ci/commit.rb
```diff
@@ -225,5 +246,13 @@ def save_yaml_error(error)
       self.yaml_errors = error
       save
     end
+
+    def status_for(scope)
+      if yaml_errors.present?
+        return 'failed'
+      end
```
nonmadden: why not?
`return 'failed' if yaml_errors.present?`

DouweM:
@dsander:

> What do you think about not showing the GenericCommitStatus in the builds tab? They can be anything, coverage reports, code quality or the build status of an external test runner.

I think the MR Builds tab should show all builds relevant to the merge request, regardless of whether those builds were run using GitLab CI or some other system, so showing `GenericCommitStatus`es is correct.

dsander:
> I think the MR Builds tab should show all builds relevant to the merge request, regardless of whether those builds were run using GitLab CI or some other system, so showing GenericCommitStatuses is correct.

@DouweM Alright, I guess then this PR will never make it.

I would have liked better communication of your (GitLab Inc.) plans, the PR got a few plus ones which lead me to continue with my approach, but it now seems that this was never an option. Communicating your idea (again the corporations) could have saved me a few hours of my time.

I hope you come up with a solution to show all the information that is exposed via the API soon.

DouweM:
@dsander Sorry about that. I didn't intend to "hijack" this feature, I simply wasn't aware of it. I wanted a Builds tab for merge requests at some point, and implemented it the same night; it wasn't part of any roadmap. Had I know about this PR, I would likely have tried to work with you to get something we could both be happy about instead of pretty much nullifying your efforts.

As for presenting all available commit status data, do you have a suggestion for how we could adapt the MR Builds tab to this effect?

dsander:
No hard feelings, after I compared the amount of MR on GitLab.com with the number here I figured it would have probably been better to open the MR at Gitlab.com.

> As for presenting all available commit status data, do you have a suggestion for how we could adapt the MR Builds tab to this effect?

Not really sure, an additional line for `CommitStatus`es table rows could show the description.

I still think the UX could be confusing, imagine GitLab MR like [this](https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/2340/builds) would also have two external statuses, finding the description of one of the statuses would already be a small effort.
If one of the external statuses failed one needs to filter out the builds that are allowed to fail until you find the one failed external status. 

Maybe I am too coined by the way GitHub summarizes the PR status, but I really like how you see the status of each service at a first glance.

DouweM:
@dsander I don't see why we should split it up by service rather than by build. 

In the case of gitlab-ce, it's made a bit more complicated because we've split the `spec` and `spinach` builds up, but really it's `spec`, `spinach`, `teaspoon`, `brakeman`, `bundler:audit`, `flay`, `flog` and `rubocop`, all of which represent very different tasks that don't make sense summarized under a single `GitLab CI status`. 

Splitting up by service only seems to make sense if a project uses Travis just for tests (`spec`, `spinach` and `teaspoon`, in our case) and different services for all of the non-spec tasks, like Hound CI, CodeClimate etc, but even then splitting up by build seems just as good.

dsander:
@DouweM I think we have a different definition of "service", for me it is not which tests are done but who executes them. As a user I know which service (GitLabCI, TravisCI, Coveralls, etc)  runs which tests, when the builds would be grouped I think it would be easier to see why the MR is marked as failure or success.

I don't like to show the external `CommitStatus`es in the builds tab because for me external statuses and internal builds are not on the same "level". This is how I think about a MR/commit and the builds/jobs of it:

```
commit/MR which triggers builds
 |- GitLabCI
    |- spec:models (Build)
    |- spec:controllers (Build)
    |- spec:benchmark allowed-to-fail (Build)
 |- TravisCI (GenericCommitStatus)
    |- spec:models (not reported to GitLab individually)
    |- spec:controllers (not reported to GitLab individually)
 |- Hound (GenericCommitStatus)
    |- rubocop (not reported to GitLab individually)
    |- eslint (not reported to GitLab individually)
    ....
 |- Coveralls (GenericCommitStatus)
    |- spec:models coverage change (not reported to GitLab individually)
    |- spec:controllers coverage change (not reported to GitLab individually)
```

The way I see it is that external services report the status for the whole MR/commit whereas GitLab(CI) reports every build, which makes sense. 
What I find confusing is that the aggregated status of external services is shown the same way like one of the possibly many GitLabCI builds which determine the GitLabCI status of the MR.

DouweM:
@dsander I agree that it's confusing that the Builds tab lists aggregated results from external services along individual GitLab CI build results, but I don't currently see a good solution. Showing GitLab CI builds as an aggregate result as well is not an option since these builds can be very diverse (like in the gitlab-ce case) and we're interested in all of their statuses. In your example, you have separate aggregate results for specs, linting and coverage, but with gitlab-ce all of these are represented by different GitLab CI builds, rather than different services.

dsander:
> Showing GitLab CI builds as an aggregate result as well is not an option since these builds can be very diverse (like in the gitlab-ce case) and we're interested in all of their statuses.

Well, I just see the aggregated statues as a quick indicator, the builds tab would still have all the details. The way I am imagining it with the usage of an accordion, the link to the build details would just be one level further down.
One could even skip the accordion as long as there are no external statues and with that keep UX for gitlab-ce as is.

DouweM:
@dsander I'm afraid I don't see enough benefit in adding a summary level between "Build failed for 123456. View details" and the Builds tab, since in the end the user is going to want to see the most detailed view anyway when finding out which build failed and why.

connorshea:

Based on the conversation above I'm not sure how likely it would be to be merged, but if you want to discuss it in an issue over at [the GitLab.com project](https://gitlab.com/gitlab-org/gitlab-ce) I'm sure we'd be happy to.

PR #9771, opened 2015-10-25
Fix: 500 error returned if destroy request without HTTP referer

kazubu:
This pull request is a fix for 500 error when gitlab received a destroy request which has no HTTP referer.

Error on logs:

```
NoMethodError (undefined method `include?' for nil:NilClass):
  app/controllers/projects_controller.rb:113:in `destroy'
```

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -12,6 +12,7 @@ v 8.2.0 (unreleased)
   - Fix: Inability to reply to code comments in the MR view, if the MR comes from a fork
   - Use git follow flag for commits page when retrieve history for file or directory
   - Show merge request CI status on merge requests index page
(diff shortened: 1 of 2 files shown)


stanhu:
Thanks! Somehow I missed this in d01fb5d81bff09. Can you take a look at those changes and use the given function?

kazubu:
Thank you for checking! I've changed the code to use redirect_back_or_default function. Could you please check it?

stanhu:
@kazubu Thanks, looks good. It looks like a conflict at the moment, probably due to the CHANGELOG entry. Can you rebase master?

kazubu:
@stanhu Thanks! I rebased it.

stanhu:
Thanks, @kazubu!

PR #9782, opened 2015-10-28
Fixed markdown issue in PROCESS.md

jvanbaarsen:
/cc @123Haynes Can you take a look?

### DIFF
--- a/PROCESS.md
+++ b/PROCESS.md
@@ -119,6 +119,6 @@ rebase with master to see if that solves the issue.
 ### Closing down the issue tracker on GitHub
 
 We are currently in the process of closing down the issue tracker on GitHub, to
-prevent duplication with the [GitLab.com issue tracker][https://gitlab.com/gitlab-org/gitlab-ce/issues].
+prevent duplication with the GitLab.com issue tracker.
 Since this is an older issue I'll be closing this for now. If you think this is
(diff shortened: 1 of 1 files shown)


PROCESS.md
```diff
@@ -119,6 +119,6 @@ rebase with master to see if that solves the issue.
 ### Closing down the issue tracker on GitHub
 
 We are currently in the process of closing down the issue tracker on GitHub, to
-prevent duplication with the [GitLab.com issue tracker][https://gitlab.com/gitlab-org/gitlab-ce/issues].
+prevent duplication with the [GitLab.com issue tracker](https://gitlab.com/gitlab-org/gitlab-ce/issues).
```
Razer6: @jvanbaarsen Please escape the braces to make the text to work with copy&paste. Right now those links are rendered has links.

jvanbaarsen: Good point!

jvanbaarsen:
@Razer6 Fixed

Razer6:
I would only use the last `GitLab.com issue tracker` as an issue.

Razer6:
It's both the same link and people would already click on the first one before continuing reading.

jvanbaarsen:
@Razer6 Hm, you I think you're right! Will update later today

Razer6:
@jvanbaarsen Ping me and I merge then

jvanbaarsen:
@Razer6 Updated (sorry for delay)

PR #9794, opened 2015-11-02
Go to gitlab installation folder before initialize database

arnaud-zg:
During installation user aren't in the gitlab folder installation.


Razer6:
Thanks

PR #9812, opened 2015-11-07
Hide tab-bar in login-box when only one tabs is shown.

kleinweby:
When an external auth provider is configured and Sign-in is disabled an tab-bar with only one tab is shown.

screen shot 2015-11-07 at 12 45 28

This change hides this tab-bar in this case.

screen shot 2015-11-07 at 12 46 13

In all other case a tab-bar is shown as before.

### DIFF
--- a/app/views/devise/shared/_signin_box.html.haml
+++ b/app/views/devise/shared/_signin_box.html.haml
@@ -7,26 +7,34 @@
       %h3 Sign in
   .login-body
     - if form_based_providers.any?
-      %ul.nav.nav-tabs
+      - if form_based_providers.count >= 2 || signin_enabled?
+        %ul.nav.nav-tabs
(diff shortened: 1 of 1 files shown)


jvanbaarsen:
Looks good to me

tsigo:
Hi @chrspeich,

This was reverted by https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/1805 due to unexpected test failures on CI.

Your idea was a good one, I think we just need to figure out a better implementation, perhaps without so much logic in the view.

jvanbaarsen:
@tsigo Just to be sure, is this still reverted?

tsigo:
@jvanbaarsen It is.

chrspeich:
Alright…

I'll look into revising this patch.

PR #9820, opened 2015-11-11
Avoid render edit_form in each notes.

huacnlee:
- Avoid render edit_form in each notes.
- Use RJS to render edit note feature.

Before:

```
Rendered projects/notes/_note.html.haml (27.9ms)
Rendered projects/_zen.html.haml (0.3ms)
Rendered projects/notes/_hints.html.haml (0.7ms)
Rendered projects/_md_preview.html.haml (3.9ms)
Rendered projects/notes/_edit_form.html.haml (6.9ms)
Rendered projects/notes/_note.html.haml (17.7ms)
Rendered projects/_zen.html.haml (0.3ms)
Rendered projects/notes/_hints.html.haml (0.6ms)
Rendered projects/_md_preview.html.haml (3.4ms)
Rendered projects/notes/_edit_form.html.haml (7.0ms)
```

After:

```
Rendered projects/notes/_note.html.haml (13.8ms)
Rendered projects/notes/_note.html.haml (7.1ms)
Rendered projects/notes/_note.html.haml (9.5ms)
Rendered projects/notes/_note.html.haml (8.5ms)
```

This change reduce at least 6ms \* N ('N' - number of notes).

### DIFF
--- a/app/controllers/projects/notes_controller.rb
+++ b/app/controllers/projects/notes_controller.rb
@@ -3,7 +3,7 @@ class Projects::NotesController < Projects::ApplicationController
   before_action :authorize_read_note!
   before_action :authorize_create_note!, only: [:create]
   before_action :authorize_admin_note!, only: [:update, :destroy]
-  before_action :find_current_user_notes, except: [:destroy, :delete_attachment]
+  before_action :find_current_user_notes, except: [:destroy, :edit, :delete_attachment]
 
   def index
     current_fetched_at = Time.now.to_i
@@ -27,6 +27,11 @@ class Projects::NotesController < Projects::ApplicationController
       format.json { render_note_json(@note) }
       format.html { redirect_back_or_default }
     end
+  end
+
+  def edit
+    @note = note
+    render layout: false
   end
 
   def update
--- a/features/steps/project/issues/issues.rb
+++ b/features/steps/project/issues/issues.rb
@@ -219,7 +219,7 @@ class Spinach::Features::ProjectIssues < Spinach::FeatureSteps
   end
 
   step 'The code block should be unchanged' do
-    expect(page).to have_content("```\nCommand [1]: /usr/local/bin/git , see [text](doc/text)\n```")
+    expect(page).to have_content("Command [1]: /usr/local/bin/git , see [text](doc/text)")
   end
 
   step 'project \'Shop\' has issue \'Bugfix1\' with description: \'Description for issue1\'' do
--- a/spec/features/task_lists_spec.rb
+++ b/spec/features/task_lists_spec.rb
@@ -51,7 +51,6 @@ feature 'Task Lists', feature: true do
 
       expect(page).to have_selector(container)
       expect(page).to have_selector("#{container} .wiki .task-list .task-list-item .task-list-item-checkbox")
-      expect(page).to have_selector("#{container} .js-task-list-field")
       expect(page).to have_selector('form.js-issuable-update')
       expect(page).to have_selector('a.btn-close')
     end
@@ -90,7 +89,6 @@ feature 'Task Lists', feature: true do
 
       expect(page).to have_selector('.note .js-task-list-container')
       expect(page).to have_selector('.note .js-task-list-container .task-list .task-list-item .task-list-item-checkbox')
-      expect(page).to have_selector('.note .js-task-list-container .js-task-list-field')
     end
 
     it 'is only editable by author' do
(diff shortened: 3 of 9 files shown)


huacnlee:
Use `Note.find(params[:id])` another reason is there has some library may use primary key for caching. 
for example: [second_level_cache](https://github.com/hooopo/second_level_cache)

huacnlee:
There have a test case not work, someone can help me? I don't know the reason

```
 ✘  Then The code block should be unchanged      # features/steps/project/issues/issues.rb:221
```

YorickPeterse:
@huacnlee I think the reason for `@project.notes.find(...)` is to ensure the note is _always_ scoped to the given project. The extra `WHERE` shouldn't really affect performance so I think it's easier to keep that query as-is.

Regarding the failing spec, I wouldn't be surprised if it checks if the edit form is present but I can't tell for certain as the CI output is massive.

Other than that I think it makes sense to only render the edit form when actually needed, instead of always rendering it.

app/controllers/projects/notes_controller.rb
```diff
@@ -61,7 +66,7 @@ def delete_attachment
   private
 
   def note
-    @note ||= @project.notes.find(params[:id])
+    @note ||= Note.find(params[:id])
```
YorickPeterse: Ah yes, I think this would now technically allow one to specify _any_ note ID in the "id" parameter and get it to render the note, thus we should keep the `@project.notes.find` form.

huacnlee: Ok, I revert that line.

huacnlee:
Ok, test passed.

huacnlee:
New change:

- Fragment caching for `_note`.
- Note action buttons visible check move into JavaScript.
- Vote status move into JavaScript.

YorickPeterse:
@randx Any comments on these changes, or is this good to go?

p.s. I don't have commit bit to this mirror so somebody else will need to merge this if it's OK to go.

randx:
@YorickPeterse thanks for mentioning me. 

@huacnlee awesome improvement. I will try now and merge if ok. Thank you!

features/steps/project/issues/issues.rb
```diff
@@ -219,7 +219,7 @@ class Spinach::Features::ProjectIssues < Spinach::FeatureSteps
   end
 
   step 'The code block should be unchanged' do
-    expect(page).to have_content("```\nCommand [1]: /usr/local/bin/git , see [text](doc/text)\n```")
+    expect(page).to have_content("Command [1]: /usr/local/bin/git , see [text](doc/text)")
```
tsigo: I don't understand this change. The point of the test was that the code block is unchanged, so why were the backticks removed?

tsigo:
I believe this change broke task lists inside notes. They're now always disabled.

spec/features/task_lists_spec.rb
```diff
@@ -51,7 +51,6 @@ def visit_issue(project, issue)
 
       expect(page).to have_selector(container)
       expect(page).to have_selector("#{container} .wiki .task-list .task-list-item .task-list-item-checkbox")
-      expect(page).to have_selector("#{container} .js-task-list-field")
```
tsigo: These three removed expectations were there for a reason -- they're required for task lists to work.

randx:
@tsigo thanks

tsigo:
@randx So should we revert this, or...?

randx:
@tsigo yes and ask for better fix

randx:
I dont think it should go in 8.2 if there is a problem with it

tsigo:
@huacnlee I've reverted this merge in https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/1804

I do like this change though, so if you can find a way to make it work without breaking task lists in notes I'd be glad to take another look.

huacnlee:
Ok, I will fix them
Robert Speicher <notifications@github.com>于2015年11月16日 周一04:40写道：

features/steps/project/issues/issues.rb
huacnlee:
```
@javascript
Scenario: Blocks inside comments should not build relative links
  Given I visit issue page "Release 0.4"
  And I leave a comment with code block
  Then The code block should be unchanged
```

```
page.within(".js-main-target-form") do
  fill_in "note[note]", with: "```\nCommand [1]: /usr/local/bin/git , see [text](doc/text)\n```"
  click_button "Add Comment"
  sleep 0.05
end
```

This action can not fail, the fill content is correct.

And that test case title is "Blocks inside comments should not build relative links", seem like means "not change links", not code block `\`\`\``

PR #9822, opened 2015-11-11
Also accept "LICENCE" as spelling for the licence file.

n0la:
Some (especially those who've learned BE) love to spell it LICENCE. As in "Licence to Kill"; but I am personally not a fan of Timothy Dalton.

### DIFF
--- a/app/models/repository.rb
+++ b/app/models/repository.rb
@@ -244,7 +244,7 @@ class Repository
   def license
     cache.fetch(:license) do
       tree(:head).blobs.find do |file|
-        file.name =~ /\Alicense/i
+        file.name =~ /\A(license|licence)/i
       end
     end
   end


axilleas:
lol :smile: 

GitHub has gone even further and you can write both the American/British way as well as mix upercase with lowercase, so eg. `LiCEnce` is valid as well :)

n0la:
The matching is case insensitive so you can write LICENCE all you like; e.g. LiCeNcE, lICEnCE. But I am not leet enough for that.

There should be a test, but my rails magick is weak and I failed miserably to find how and where I'd have to write a test for it.

axilleas:
@n0la nah, don't bother with the regex, I just mentioned it for reference.

@stanhu @haynes can you help with the tests? :)

stanhu:
I think this PR could be folded into [this MR](https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/1361) might be relevant.

/cc: @ZJvandeWeg

ZJvandeWeg:
Will add this now. Should be in the MR stanhu just referenced in an hour.

stanhu:
Thanks, @ZJvandeWeg! Closing this PR in favor of yours.

PR #9825, opened 2015-11-12
Add caching for ApplicationSetting, Ci::ApplicationSetting.

huacnlee:
ApplicationSetting.current was called in every pages, cache it and expires it after it updated.

This changes will avoid a SQL query in every pages 

``` SQL
SELECT  "application_settings".* FROM "application_settings"   ORDER BY "application_settings"."id" DESC LIMIT 1
```

### DIFF
--- a/app/models/application_setting.rb
+++ b/app/models/application_setting.rb
@@ -68,8 +68,14 @@ class ApplicationSetting < ActiveRecord::Base
     end
   end
 
+  after_commit do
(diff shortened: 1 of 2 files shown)


randx:
cc @YorickPeterse

randx:
@huacnlee thank you. Looks good

huacnlee:
:smile:

PR #9827, opened 2015-11-12
Add migration to change web_hooks url field to text type from string.

adreyer:
Before this commit The webhook url was a string field which is limited
by default to 255 characters. In some cases this might be too short for
the desired webhook url. After this it is changed to a text field which
is unbounded to allow urls of any length.


adreyer:
This should fix:
https://gitlab.com/gitlab-org/gitlab-ce/issues/1917
and is duplicated:
https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/2289

rymai:
Thanks @adreyer. Merge request https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/1940 was merged since then.

PR #9838, opened 2015-11-17
Fix github importer to handle empty issues

evindor:
In our setup this led to project import failing.

### DIFF
--- a/lib/gitlab/github_import/importer.rb
+++ b/lib/gitlab/github_import/importer.rb
@@ -19,7 +19,7 @@ module Gitlab
           if issue.pull_request.nil?
 
             body = @formatter.author_line(issue.user.login)
-            body += issue.body
+            body += issue.body || ""
 
             if issue.comments > 0
               body += @formatter.comments_header


evindor:
A bit more detail, log:
```
2015-11-17_17:36:31.47788 2015-11-17T17:36:31.477Z 3158 TID-gorsk RepositoryImportWorker JID-d144cdcdc393f15b41a95757 INFO: start
2015-11-17_17:36:34.59514 2015-11-17T17:36:34.595Z 3158 TID-gorsk RepositoryImportWorker JID-d144cdcdc393f15b41a95757 INFO: fail: 3.118 sec
2015-11-17_17:36:34.59661 2015-11-17T17:36:34.596Z 3158 TID-gorsk WARN: {"retry"=>true, "queue"=>"gitlab_shell", "class"=>"RepositoryImportWorker", "args"=>[12], "jid"=>"d144cdcdc393f15b41a95757", "enqueued_at"=>1447781791.471111, "error_message"=>"no implicit conversion of nil into String", "error_class"=>"TypeError", "failed_at"=>1447781794.5944014, "retry_count"=>0}
2015-11-17_17:36:34.59679 2015-11-17T17:36:34.596Z 3158 TID-gorsk WARN: no implicit conversion of nil into String
2015-11-17_17:36:34.59699 2015-11-17T17:36:34.596Z 3158 TID-gorsk WARN: /opt/gitlab/embedded/service/gitlab-rails/lib/gitlab/github_import/importer.rb:22:in `+'
```

After this tiny fix the import succeeded.

Empty-body issues are common when using waffle.

stanhu:
Looks good to me, although the [GitHub API documention](https://developer.github.com/v3/issues/
) doesn't say that `null` is a valid value here, although it does say this explicitly in other fields. I would be concerned that we would need to do `nil` checks everywhere for strings that are empty.

Is waffle creating these issues?

evindor:
@stanhu Yes, in waffle when you create a card it's an issue in the connected repo - with title set from waffle and an empty body.

Regarding `nil` checks - the api docs page doesn't explicitly say anything about empty body value :confused:

lib/gitlab/github_import/importer.rb
```diff
@@ -19,7 +19,7 @@ def execute
           if issue.pull_request.nil?
 
             body = @formatter.author_line(issue.user.login)
-            body += issue.body
+            body += issue.body.nil? ? '' : issue.body
```
DouweM: `issue.body || ""` please!

evindor: Updated!

DouweM:
Thanks @evindor!

PR #9846, opened 2015-11-22
Update autocomplete_controller to be more readable

felipeorlando:
This turn autocomplete controller more readable. It's the only reason for change, but it's unnecessary use 'unless' statement.


randx:
Thanks

Issue #9872, opened 2015-12-01
Impersonating blocked user logs you off


Baertierchen:
When trying to impersonate a user that is currently blocked, you end up being logged off. After logging back in with your own credentials, the button to stop impersonation appears even though you are logged in as yourself at this time.
Would be nicer if you were simply told that you can't impersonate blocked users and remain logged in.
That's on 8.2.1 tag.

atomaka:
How about #9873 to fix this?  Along with alerting when a post request is made to impersonate a user, it prevents display of the Impersonate button altogether when a user is blocked.

stanhu:
#9873 seems like a sensible approach.

PR #9873, opened 2015-12-02
Prevent impersonation if blocked

atomaka:
Issue #9872 notes that impersonating a blocked user will log you out of your account. A more desirable result is for the administrator to be notified that a blocked user cannot be impersonated.  This pull request addresses that in two ways:

When viewing a users page in the administration panel, the Impersonate button will not be present when the user is blocked:

screen shot 2015-12-02 at 8 11 18 am

If an attempt to post against the impersonate URL for a blocked user occurs, a flash message will display notifying the administrator that a blocked user cannot be impersonated:

screen shot 2015-12-02 at 8 14 58 am

Test cases are included for both of these changes.

I have also included a minor change that explicitly defines the `impersonator_return_to` location instead of relying on `HTTP_REFERER`.

### DIFF
--- a/app/controllers/admin/impersonation_controller.rb
+++ b/app/controllers/admin/impersonation_controller.rb
@@ -5,14 +5,20 @@ class Admin::ImpersonationController < Admin::ApplicationController
   before_action :authorize_impersonator!
 
   def create
-    session[:impersonator_id] = current_user.username
-    session[:impersonator_return_to] = request.env['HTTP_REFERER']
+    if @user.blocked?
+      flash[:alert] = "You cannot impersonate a blocked user"
 
-    warden.set_user(user, scope: 'user')
+      redirect_to admin_user_path(@user)
+    else
+      session[:impersonator_id] = current_user.username
+      session[:impersonator_return_to] = request.env['HTTP_REFERER']
 
-    flash[:alert] = "You are impersonating #{user.username}."
+      warden.set_user(user, scope: 'user')
 
-    redirect_to root_path
+      flash[:alert] = "You are impersonating #{user.username}."
+
+      redirect_to root_path
+    end
   end
 
   def destroy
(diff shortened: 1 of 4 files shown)


atomaka:
I cannot recreate the test failure locally.

app/controllers/admin/impersonation_controller.rb
```diff
@@ -5,14 +5,20 @@ class Admin::ImpersonationController < Admin::ApplicationController
...
-    session[:impersonator_id] = current_user.username
-    session[:impersonator_return_to] = request.env['HTTP_REFERER']
+    if @user.blocked?
+      flash[:alert] = "You cannot impersonate a blocked user"
 
-    warden.set_user(user, scope: 'user')
+      redirect_to admin_user_path(@user)
+    else
+      session[:impersonator_id] = current_user.username
+      session[:impersonator_return_to] = request.env['HTTP_REFERER']
```
stanhu: I know this is an existing issue, but it's possible `HTTP_REFERER` is not sent by the client. Is there a default URL to fall back to if this doesn't exist?

stanhu:
Thanks! Looks good. Triggered a rebuild.

app/controllers/admin/impersonation_controller.rb
atomaka: Perhaps `admin_user_path(@user)` could replace `HTTP_REFERER` completely?  I believe that's the only intended outcome and the tests continue to pass with that change.

atomaka:
I did see this in another run of the test suite (1 out of 3).  It's not clear to me how my changes could have caused this occasional failure.

app/controllers/admin/impersonation_controller.rb
stanhu: Sounds good to me.

stanhu:
@atomaka The build failures are irrelevant--false positives are unfortunately a problem at the moment.

stanhu:
:+1: Thanks, @atomaka!

Issue #9874, opened 2015-12-02
User should be informed when personal project limit reached


vellamike:
When the personal project limit on gitlab is reached the '+' symbol for a new project simply disappears. The + symbol should remain and user should be informed that they can't add a new project and need to inform their admin. Otherwise it's pretty confusing to understand why you can't add a new personal project.

atomaka:
I've implemented something close to your suggestion in #9876, @vellamike.  I'm not much of a designer so I'd love suggestions!

atomaka:
This was resolved in #9876, @stanhu.

Razer6:
Thanks

PR #9876, opened 2015-12-03
Clarify Project Limit Reached

atomaka:
Issue #9874 points at that realizing your project limit has been reached can be difficult.  Ideally, the user needs to be notified in some way.  This pull request adds an alert to the user's dashboard when they have reached their project limit.

project-limit

This functions the same way as the alerts for when a user does not have an SSH key or a password and can be hidden temporarily or permanently.  This is accomplished using Javascript to create a cookie when the user clicks `Remind later` and a boolean field on the `User` model (`hide_project_limit`) when the user clicks `Don't show again`.  Based on the truthiness of these two values and the status of `User.can_create_project?` the message will be displayed or not.

If accepted, a follow up pull request will need to be issued to address the repetitive code.  This will be the third time that similar code is used and it should be abstracted.  I thought adding that to this pull request would be too much, but could visit it now if you'd like.

### DIFF
--- a/app/assets/stylesheets/pages/projects.scss
+++ b/app/assets/stylesheets/pages/projects.scss
@@ -5,7 +5,7 @@
     font-weight: normal;
   }
 }
-.no-ssh-key-message {
+.no-ssh-key-message, .project-limit-message {
   background-color: #f28d35;
   margin-bottom: 16px;
 }
--- a/app/assets/javascripts/user.js.coffee
+++ b/app/assets/javascripts/user.js.coffee
@@ -2,3 +2,9 @@ class @User
   constructor: ->
     $('.profile-groups-avatars').tooltip("placement": "top")
     new ProjectsList()
+
+    $('.hide-project-limit-message').on 'click', (e) ->
+      path = '/'
+      $.cookie('hide_project_limit_message', 'false', { path: path })
+      $(@).parents('.project-limit-message').remove()
+      e.preventDefault()
--- a/app/controllers/profiles_controller.rb
+++ b/app/controllers/profiles_controller.rb
@@ -70,6 +70,7 @@ class ProfilesController < Profiles::ApplicationController
       :email,
       :hide_no_password,
       :hide_no_ssh_key,
+      :hide_project_limit,
       :linkedin,
       :location,
       :name,
--- a/app/views/dashboard/_projects_head.html.haml
+++ b/app/views/dashboard/_projects_head.html.haml
@@ -1,3 +1,6 @@
+= content_for :flash_message do
+  = render 'shared/project_limit'
+
 %ul.center-top-menu
   = nav_link(path: ['projects#index', 'root#index']) do
     = link_to dashboard_projects_path, title: 'Home', class: 'shortcuts-activity', data: {placement: 'right'} do
--- a/app/views/shared/_project_limit.html.haml
+++ b/app/views/shared/_project_limit.html.haml
@@ -1,0 +1,8 @@
+- if cookies[:hide_project_limit_message].blank? && !current_user.hide_project_limit && !current_user.can_create_project?
+  .project-limit-message.alert.alert-warning.hidden-xs
+    You won't be able to create new projects because you have reached your project limit.
+
+    .pull-right
+      = link_to "Don't show again", profile_path(user: {hide_project_limit: true}), method: :put, class: 'alert-link'
+      |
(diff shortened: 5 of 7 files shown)


stanhu:
Thanks for doing this! The strikethrough on the button looks a little funny to me. Perhaps it should just be grayed out?

stanhu:
Alternatively, should the button still be available but a flash message be displayed if the user is over the limit? Thoughts, @creamzy?

atomaka:
That was my initial thought, @stanhu.  I thought an implementation similar to the old SSH Key message might be reasonable.  But that was removed sometime recently so I wasn't sure if re-adding a message that sticky was something that people were interested in.

atomaka:
I also agree that the strike through is not very appealing.  I added that when I was still using the plus icon because it still looked like it might be an active button.  It might be more reasonable now. Or maybe it would be better to replace the cursor with the icon similar to how some of the buttons work after press.

Razer6:
@atomaka What about the top-right botton near the search bar? What about the API? I think this also needs backend changes.

atomaka:
@Razer6 Gitlab currently prevents project creation.  It's just difficult for the user to understand why they do not have the buttons that they used to.  The API definitely [accounts for and returns an appropriate response](https://github.com/gitlabhq/gitlabhq/blob/master/lib/api/projects.rb#L107) in this case.

The button near the search bar is an excellent point, though, and perhaps @stanhu's suggestion of an alert is a better option.

atomaka:
I lied, the SSH key warning was not removed.  Guess I don't run around without an SSH key very often.

I like the idea of this for warning users about not having projects.

ssh-warning

It should be cleaner code and and easier to test, too.

atomaka:
The latest commit implements this functionality as a flash message following the same pattern as the message when you do not have an SSH commit added.  Currently, I've added display only to the dashboard:

project-flash

app/assets/javascripts/application.js.coffee
```diff
@@ -197,3 +197,9 @@ $ ->
     new ConfirmDangerModal(form, text)
 
   new Aside()
+
+  $('.hide-project-limit-message').on 'click', (e) ->
+    path = '/'
+    $.cookie('hide_project_limit_message', 'false', { path: path })
+    $(@).parents('.project-limit-message').remove()
+    e.preventDefault()
```
atomaka: This code obviously doesn't belong in `application.js.coffee` but I wasn't sure where it should go that it would actually be executed.

app/assets/stylesheets/pages/projects.scss
```diff
@@ -9,6 +9,10 @@
   background-color: #f28d35;
   margin-bottom: 16px;
 }
+.project-limit-message {
+  background-color: #f28d35;
+  margin-bottom: 16px;
+}
```
atomaka: Copy pasted from above.  Maybe shared style?

atomaka:
PR description updated for latest commit.

app/assets/javascripts/application.js.coffee
stanhu: I was put this in a partial, just like `_no_ssh.html.haml`, and render it anywhere the "New Project" button shows up.

app/assets/stylesheets/pages/projects.scss
stanhu: Yes, I'd suggest either using SCSS inheritance or defining some variables.

app/assets/javascripts/application.js.coffee
atomaka: The new code is rendered in partial `_project_limit_message.html.haml`.  I'm specifically talking about placement of the Javascript.  Currently, similar Javascript for the no ssh key message resides in [project.js.coffee](https://github.com/gitlabhq/gitlabhq/blob/master/app/assets/javascripts/project.js.coffee#L22).  However, this code doesn't seem to execute on the dashboard. Unfortunately, I'm not entirely sure how Javascript execution is determined in GitLab so I wasn't sure which file was best suited for this piece of code and shoved it in `application.js.coffee`.

atomaka:
I adjusted the location for the Javascript and made a minor change to the CSS.

stanhu:
Looks good to me. @randx, what do you think?

randx:
@stanhu looks good to me. Merge if CI green

atomaka:
Rebase against master fixed the two failing tests locally.

atomaka:
@stanhu All green!

stanhu:
Thanks, @atomaka!

PR #9882, opened 2015-12-04
update nginx config hint

darkrasid:
When use gitlab with relative url, 
user should change location regular expression on nginx config file.

Because example of nginx config file contain this type regular expression.

```
^/[\w\.-]+/[\w\.-]+/(info/refs|git-upload-pack|git-receive-pack)$
```

it will catch some URL like below.

```
/project/repo.git/git-upload-pack
```

However if user set relative url, upper url should be

```
/relative_url/project/repo.git/git-upload-pack
```

so regular expression isn't able to catch.

It makes errors when cloing, pulling, pushing or archiving.  
(ex. it will clone empty repository, even if the repository does not empty.)

I think it is better way that comment to user in example of nginx config file

### DIFF
--- a/lib/support/nginx/gitlab
+++ b/lib/support/nginx/gitlab
@@ -114,6 +114,9 @@ server {
   }
 
   location ~ ^/[\w\.-]+/[\w\.-]+/gitlab-lfs/objects {
+    ## If you use relative url, please uncomment below line and remove origin line.
+    # location ~ ^/YOUR_RELATIVE_URL/[\w\.-]+/[\w\.-]+/gitlab-lfs/objects {
+
     client_max_body_size 0;
     # 'Error' 418 is a hack to re-use the @gitlab-workhorse block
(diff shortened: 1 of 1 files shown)


lib/support/nginx/gitlab
```diff
@@ -114,6 +114,9 @@ server {
   }
 
   location ~ ^/[\w\.-]+/[\w\.-]+/gitlab-lfs/objects {
+    ## If you use relative url, please uncomment below line and remove origin line.
+    # location ~ ^/YOUR_RELATIVE_URL/[\w\.-]+/[\w\.-]+/gitlab-lfs/objects {
```
jacobvosmaer: We should not repeat the entire regex in the comment, that is a maintenance hassle. We should rather say something like "change `^/` to `^/gitlab/`.

darkrasid:
I understand you point and I change it.

However I take wrong branch on this PR so I make another one and close this one.
Please check here #9894

Issue #9035, opened 2015-03-26
Preserve trailing new lines at the end of file


chulkilee:
Related: #5936

- GitLab 7.9.1
- Firefox 36.0.4 on Mac

When I create or edit a text file via web editor, it removes the last new line. For example, when I put three new blank lines, only two lines are saved.

kylekatarnls:
This works well for me: #9422

Razer6:
Closing as seems to be fixed

eddiemonge:
This isn't fixed yet as that PR is still open. Can this stay open until then? Also fine with keeping it closed as a duplicate of #8706

Razer6:
Thanks for clarification. Then leave it closed as a duplicate of #8706 because this one was open before.

PR #9422, opened 2015-06-30
#9035 + #5936 Preserve trailing new lines at the end of file

kylekatarnls:
If the content end with a new line, add it as raw

PR #9884, opened 2015-12-04
#9035 + #5936 Rebase

kylekatarnls:
Rebase the patch branch

### DIFF
--- a/app/views/projects/blob/_editor.html.haml
+++ b/app/views/projects/blob/_editor.html.haml
@@ -17,7 +17,10 @@
 
   .file-content.code
     %pre.js-edit-mode-pane#editor
-      = params[:content] || local_assigns[:blob_data]
+      - content = params[:content] || local_assigns[:blob_data]
+      = content
+      - if content.length && 0 === content[content.length-1].index(/[\r\n]/)
+        != "\n"
     - if local_assigns[:path]
       .js-edit-mode-pane#preview.hide
         .center


app/views/projects/blob/_editor.html.haml
```diff
@@ -17,7 +17,10 @@
 
   .file-content.code
     %pre.js-edit-mode-pane#editor
-      = params[:content] || local_assigns[:blob_data]
+      - content = params[:content] || local_assigns[:blob_data]
+      = content
+      - if content.length && 0 === content[content.length-1].index(/[\r\n]/)
```
stanhu: I'm a bit confused how this works. There's an `if` clause here with a non-printing line.

kylekatarnls: It do what this issues are about: preserve the trailing end of line (avoid the auto-trim in previous version and now avoid a line to be added automatically). In short word, you will not have the end of the file modified if you haven't added manually: if the last char is a new line, re-add it: https://github.com/gitlabhq/gitlabhq/pull/9422

stanhu: I did some cursory debugging and found the right number of newlines get sent from the UI editor to the controller. Is the issue that the view is truncating the last newline if it's there?

kylekatarnls: No, the other way, the last new line from the controller to the editor was lost.

dosire:
This problem is solved in master.

dosire:
See https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/2099

PR #9855, opened 2015-11-25
Add new feature to find file

koreamic:
Using the regular expression, develop "file finder" feature.
(http://feedback.gitlab.com/forums/176466-general/suggestions/4987909-add-file-finder-fuzzy-input-in-files-tab-to-ju)

Issue(#9854)

Issue #9854, opened 2015-11-25
File Finder Feature


koreamic:
I've been using Gitlab and I think it is awesome!!!.

As I know, Gitlab has no 'file finder' feature like Github
(https://github.com/blog/793-introducing-the-file-finder)

I read the following old article in your feedback.
(http://feedback.gitlab.com/forums/176466-general/suggestions/4987909-add-file-finder-fuzzy-input-in-files-tab-to-ju)

Using not the fuzzy logic but the regular expression, I developed the feature.

If I create a pull request, review my pull request?

yms9654:
I think It is a very convenient feature.
I can find a file quick and instantly.

stanhu:
@koreamic Yes, please submit a PR for this!

PR #9889, opened 2015-12-05
Add new feature to find file(recreate)

koreamic:
Using the fuzzy lib, develop "file finder" feature.
(http://feedback.gitlab.com/forums/176466-general/suggestions/4987909-add-file-finder-fuzzy-input-in-files-tab-to-ju)

Issue(#9854)
PR(#9855)
- apply ajax request
- fuzzy filter(https://github.com/jeancroy/fuzzaldrin-plus)
- short cut(when "t" was hitted, go to 'file find' page and 'esc' is to go back)
- feature screenshot
  file_finder

(ps. cause by my mistake, recreate PR(#9855), sorry!!)


stanhu:
Great work! I assume since this depends on [this gitlab_git MR](https://gitlab.com/gitlab-org/gitlab_git/merge_requests/60), this PR has to wait until a new version is release?

Can you add tests for this?

app/controllers/projects/find_file_controller.rb
```diff
@@ -0,0 +1,26 @@
+# Controller for viewing a repository's file structure
+class Projects::FindFileController < Projects::ApplicationController
+  include ExtractsPath
+  include ActionView::Helpers::SanitizeHelper
+  include TreeHelper
+  before_action :assign_ref_vars
```
stanhu: Do we also need to add:

```ruby
  before_action :require_non_empty_project
  before_action :authorize_download_code!
```

stanhu:
Does this PR also need an upgrade of gitlab_git to 7.2.22?

stanhu:
Also, this feature could use a spinach test to test the JavaScript functionality.

koreamic:
@stanhu
Thank you for your concern and help.

Yes. This PR  needs an upgrade of gitlab_git to 7.2.22.

and okay I will check to use a spinach test.

If you need to additional works. please comment.

Thank you. have a nice day.

app/assets/javascripts/shortcuts_navigation.coffee
```diff
@@ -14,6 +14,7 @@ class @ShortcutsNavigation extends Shortcuts
     Mousetrap.bind('g m', -> ShortcutsNavigation.findAndFollowLink('.shortcuts-merge_requests'))
     Mousetrap.bind('g w', -> ShortcutsNavigation.findAndFollowLink('.shortcuts-wiki'))
     Mousetrap.bind('g s', -> ShortcutsNavigation.findAndFollowLink('.shortcuts-snippets'))
+    Mousetrap.bind('t', -> ShortcutsNavigation.findAndFollowLink('.shortcuts-find-file'))
```
Razer6: Shouldn't this only be active to on the tree page? Can you also update https://github.com/gitlabhq/gitlabhq/blob/master/app/views/help/_shortcuts.html.haml and the screenshot in https://github.com/gitlabhq/gitlabhq/blob/master/doc/workflow/shortcuts.md

koreamic: Okay, @Razer6 Thank you for your suggestions.
I updated ```_shortcuts.html.haml``` and screenshot.
and I added shortcuts using up and down array key in find file view to select file more conveniently.

The reason that add ```t``` shortcut in ```app/assets/javascripts/shortcuts_navigation.coffee``` is to use find file view other page(eg. merge request/commit/branch page) 
If necessary.

but If not necessary, I just define it on the tree view.

what do you think about that?

koreamic:
@stanhu 
I added spinach test case and shortcuts that up down arrow to select item easly.
and modified some bugs(main menu bar out focusing)

please check and review sources

thank you.

yms9654:
:+1: This is a great feature!
I think many users like this.

app/assets/javascripts/lib/fuzzaldrin-plus.js
```diff
@@ -0,0 +1,1158 @@
+(function e(t,n,r){function s(o,u){if(!n[o]){if(!t[o]){var a=typeof require=="function"&&require;if(!u&&a)return a(o,!0);if(i)return i(o,!0);var f=new Error("Cannot find module '"+o+"'");throw f.code="MODULE_NOT_FOUND",f}var l=n[o]={exports:{}};t[o][0].call(l.exports,function(e){var n=t[o][1][e];return s(n?n:e)},l,l.exports,e,t,n,r)}return n[o].exports}var i=typeof require=="function"&&require;for(var o=0;o<r.length;o++)s(r[o]);return s})({1:[function(require,module,exports){
```
randx: Please move this file to `/vendor/assets/javascripts`

randx:
Please provide a screenshots for this feature

app/assets/javascripts/project_find_file.js.coffee
```diff
@@ -0,0 +1,110 @@
...
+    target = txt.toString()
+    for char in charArr
+      charIdx = target.toLowerCase().indexOf char.toLowerCase()
+      result += "#{target.substring 0, charIdx}#{(target.charAt charIdx).bold()}"
+      target = target.substring charIdx + 1
+    return (result + target)
+
+  # make tbody row html
+  makeHtml: (filePath, blobItemUrl) ->
+    return  @row.replace("%blobItemUrl", blobItemUrl).replace("%filePath", filePath)
```
stanhu: What happens if the file path has the text `%blobItemUrl`? Can we just construct the entire text without a `replace`?

app/assets/javascripts/project_find_file.js.coffee
```diff
@@ -0,0 +1,110 @@
...
+
+  # go back page
+  goBack: (event) ->
+    if event.keyCode == 27
+      event.preventdefault
```
stanhu: Isn't this `event.preventDefault()`?

app/assets/javascripts/project_find_file.js.coffee
```diff
@@ -0,0 +1,110 @@
...
+      parent.history.back()
+    return false;
+
+  # select row and focus
+  selecteRow: (rows, seletedRow, which) ->
```
stanhu: Is this a misspelling: `selectedRow`?

app/assets/javascripts/project_find_file.js.coffee
```diff
@@ -0,0 +1,110 @@
...
+    # unbind and bind keyup/keydown event at text input box
+    $(document).off "keyup"
+    $(document).off "keyup", @element
+    $(document).off "keydown", "body"
+
+    $(document).on "keyup", @goBack
+    $(document).on "keyup", @element, (event) =>
+      keyCode = event.keyCode if event
+      # up down right left arrow(37~40) esc(27) caps(20) ctrl(17) alt(18) shift(16) enter(13) cmd(91)
+      if (keyCode < 37 || keyCode > 40) && keyCode != 20 && keyCode != 17 && keyCode != 18 && keyCode != 16 && keyCode != 13 && keyCode != 91
```
stanhu: Should we define these key codes as constants (e.g. https://github.com/cybertoothca/keyevent/blob/master/src/keyevent.js)?

app/assets/javascripts/project_find_file.js.coffee
```diff
@@ -0,0 +1,109 @@
+class @ProjectFindFile
+  constructor: (@options)->
+    @filePaths = {}
+    @element = ".file-finder-input"
+    @row = "<tr class='tree-item'><td class='tree-item-file-name'><i class='fa fa-file-text-o fa-fw'></i><span class='str-truncated'><a href=%blobItemUrl>%filePath</a></span></td></tr>"
```
DouweM: Please build this structure using jQuery. Doing it as a string opens us up to XSS.

app/assets/javascripts/project_find_file.js.coffee
```diff
@@ -0,0 +1,109 @@
...
+    # focus text input box
+    $(@element).focus()
+
+    # load file list
+    @load(@options.url)
+
+  # init key event
+  initKeyEvent: ->
+    # unbind and bind keyup/keydown event at text input box
+    $(document).off "keyup"
```
DouweM: I don't think we want to get rid of all keyup events. Can you namespace these to be specific to `ProjectFindFile`? 

From https://api.jquery.com/bind/:

> If the eventType string contains a period (.) character, then the event is namespaced. The period character separates the event from its namespace. For example, in the call .bind( "click.name", handler ), the string click is the event type, and the string name is the namespace. Namespacing allows us to unbind or trigger some events of a type without affecting others. See the discussion of .unbind() for more information.

app/assets/javascripts/project_find_file.js.coffee
```diff
@@ -0,0 +1,109 @@
...
+    $(document).off "keyup", @element
+    $(document).off "keydown", "body"
+
+    # $(document).on "keyup", @goBack
+    $(document).on "keyup", (event) =>
+      if event.keyCode == 27
+        event.preventdefault
```
DouweM: `event.preventDefault()`

app/assets/javascripts/project_find_file.js.coffee
```diff
@@ -0,0 +1,109 @@
...
+        path = $('.tree-item.selected .tree-item-file-name a').attr('href')
+        if path then location.href = path
+
+  # find file
+  findFile: (searchTxt) ->
+      result = if searchTxt then fuzzaldrinPlus.filter(@filePaths, searchTxt) else @filePaths
+      @renderList result, searchTxt
+
+  # select row and focus
+  selecteRow: (rows, seletedRow, which) ->
```
DouweM: - `selectRow`
- `selectedRow`

app/assets/javascripts/project_find_file.js.coffee
```diff
@@ -0,0 +1,109 @@
...
+        @selecteRow $("tr.tree-item")
+
+    $(document).on "keydown", "body", (event) =>
+      keyCode = event.keyCode if event
+      if $("#search:focus").length > 0 && (keyCode == 38 || keyCode == 40 || keyCode == 13)
+        return false
+
+      # 38 up 40 down
+      if keyCode == 38 || keyCode == 40
+          @selecteRow $("tr.tree-item"), $('.tree-item.selected'), keyCode
```
DouweM: Inconsistent indentation

app/assets/javascripts/project_find_file.js.coffee
```diff
@@ -0,0 +1,109 @@
...
+          return false
+      else if keyCode is 13
+        path = $('.tree-item.selected .tree-item-file-name a').attr('href')
+        if path then location.href = path
+
+  # find file
+  findFile: (searchTxt) ->
+      result = if searchTxt then fuzzaldrinPlus.filter(@filePaths, searchTxt) else @filePaths
```
DouweM: Inconsistent indentation

app/assets/javascripts/project_find_file.js.coffee
```diff
@@ -0,0 +1,109 @@
...
+    $(document).on "keyup", (event) =>
+      if event.keyCode == 27
+        event.preventdefault
+        location.href = @options.treeUrl
+      return false
+
+    $(document).on "keyup", @element, (event) =>
+      keyCode = event.keyCode if event
+      # up down right left arrow(37~40) esc(27) caps(20) ctrl(17) alt(18) shift(16) enter(13) cmd(91)
+      if (keyCode < 37 || keyCode > 40) && keyCode != 20 && keyCode != 17 && keyCode != 18 && keyCode != 16 && keyCode != 13 && keyCode != 91
```
DouweM: Why are we checking all these keycodes?

app/assets/javascripts/project_find_file.js.coffee
```diff
@@ -0,0 +1,109 @@
...
+
+    $(document).on "keydown", "body", (event) =>
+      keyCode = event.keyCode if event
+      if $("#search:focus").length > 0 && (keyCode == 38 || keyCode == 40 || keyCode == 13)
+        return false
+
+      # 38 up 40 down
+      if keyCode == 38 || keyCode == 40
+          @selecteRow $("tr.tree-item"), $('.tree-item.selected'), keyCode
+          return false
```
DouweM: No need to return since there are no statements after this if/else.

app/assets/javascripts/project_find_file.js.coffee
```diff
@@ -0,0 +1,109 @@
...
+      seletedRow.addClass("selected").focus()
+
+  # files pathes load
+  load: (url) ->
+    $.ajax
+      url: url
+      method: "get"
+      dataType: "json"
+      success: $.proxy((data) ->
+        $(".loading").hide()
```
DouweM: We need to make this more specific in case we get other `.loading`s in the future.

app/assets/javascripts/project_find_file.js.coffee
```diff
@@ -0,0 +1,109 @@
...
+        seletedRow = rows.eq(0)
+      seletedRow.addClass("selected").focus()
+
+  # files pathes load
+  load: (url) ->
+    $.ajax
+      url: url
+      method: "get"
+      dataType: "json"
+      success: $.proxy((data) ->
```
DouweM: Why do we need $.proxy?

app/assets/javascripts/project_find_file.js.coffee
```diff
@@ -0,0 +1,109 @@
...
+        $(".loading").hide()
+        @filePaths = data
+        @findFile($(@element).val())
+        @selecteRow $("tr.tree-item")
```
DouweM: We need to make all of these more specific :) Can we pass an element to `new ProjectFindFile`, and then use `@element.find('tr.tree-item')` etc?

app/assets/javascripts/project_find_file.js.coffee
```diff
@@ -0,0 +1,109 @@
...
+      , this)
+
+  # render result
+  renderList: (filePaths, searchTxt) ->
+    $(".tree-table > tbody").empty()
+    for filePath, i in filePaths
+      break if i > 20
```
DouweM: This means we break when `i == 21`, which means we had 0..20 before this: 21 renders :) We need only 20.

app/assets/javascripts/project_find_file.js.coffee
```diff
@@ -0,0 +1,109 @@
...
+  # find file
+  findFile: (searchTxt) ->
+      result = if searchTxt then fuzzaldrinPlus.filter(@filePaths, searchTxt) else @filePaths
+      @renderList result, searchTxt
+
+  # select row and focus
+  selecteRow: (rows, seletedRow, which) ->
+    if rows && rows.length > 0
+      if seletedRow && seletedRow.length > 0
+        next = null
```
DouweM: Why is this necessary?

app/assets/javascripts/project_find_file.js.coffee
```diff
@@ -0,0 +1,109 @@
...
+    $(".tree-table > tbody").empty()
+    for filePath, i in filePaths
+      break if i > 20
+      markedFilePath = if searchTxt then highlightText filePath, searchTxt.split("") else filePath
+      blobItemUrl = "#{@options.blobUrlTemplate}/#{filePath}"
+      html = @makeHtml markedFilePath, blobItemUrl
+      $(".tree-table > tbody").append(html)
+
+  # highlight text(awefwbwgtc -> <b>a</b>wefw<b>b</b>wgt<b>c</b> )
+  highlightText = (txt, charArr) ->
```
DouweM: This is the highlighting code that the Atom editor uses: https://github.com/atom/fuzzy-finder/blob/master/lib%2Ffuzzy-finder-view.coffee#L70-L88 Maybe we can use the same logic :)

app/assets/javascripts/project_find_file.js.coffee
DouweM: As mentioned above, please generate the row DOM using jQuery.

app/assets/javascripts/shortcuts_navigation.coffee
DouweM: I think it doesn't make sense on the issues page, but it does on any file pages (tree, blob, blame) as well as the project root.

app/controllers/projects/find_file_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+  before_action :require_non_empty_project
+  before_action :assign_ref_vars
+  before_action :authorize_download_code!
+
+  def show
+    return render_404 unless @repository.commit(@ref)
+
+    @url = namespace_project_files_path(@project.namespace, @project, @ref, @options.merge(format: :json))
+    @treeUrl = namespace_project_tree_path(@project.namespace, @project, @ref)
+    @blobUrlTemplate = namespace_project_blob_path(project.namespace, project, tree_join(@id || @commit.id))
```
DouweM: We don't need all of these in the controller, we can just use the helpers in the view.

app/controllers/projects/find_file_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+
+    respond_to do |format|
+      format.html
+    end
+  end
+
+  def list
+    filePaths = @repo.ls_files(@ref)
```
DouweM: Ruby wants `snake_case`, not `camelCase` :)

app/views/projects/_find_file_link.html.haml
```diff
@@ -0,0 +1,6 @@
+- if project && project_namespace && ref
+  .find-file-link-holder
+    %span.btn-group{class: 'btn-group pull-right hidden-xs hidden-sm'}
+      = link_to namespace_project_find_file_path(project_namespace, project, ref), class: 'btn btn-success shortcuts-find-file', rel: 'nofollow' do
+        = icon('search')
+        %span FIND FILE
```
DouweM: This can be `Find File`, it will automatically be rendered all caps :)

app/views/projects/_find_file_link.html.haml
```diff
@@ -0,0 +1,6 @@
+- if project && project_namespace && ref
+  .find-file-link-holder
+    %span.btn-group{class: 'btn-group pull-right hidden-xs hidden-sm'}
+      = link_to namespace_project_find_file_path(project_namespace, project, ref), class: 'btn btn-success shortcuts-find-file', rel: 'nofollow' do
```
DouweM: I don't think we need `btn-success` here.

app/views/projects/_find_file_link.html.haml
```diff
@@ -0,0 +1,6 @@
+- if project && project_namespace && ref
```
DouweM: We can just use `@project` and `@ref` directly here, and we don't need to verify that they exist since they always do on the tree page :)

DouweM:
@koreamic What happened to Network shortcuts in the shortcuts screenshot? :)

app/assets/javascripts/project_find_file.js.coffee
Razer6: We already use moustrap (https://github.com/ccampbell/mousetrap), is this keycode checking necessary with that?

koreamic:
Hi. Happy new year.

@DouweM , @stanhu , @randx , @Razer6 

I updated sources for this feature. and I accepted most of your advice.

- updated shortcuts screenshot
- used mousetrap for shortcuts(up/down/esc/enter, avoid to bind and unbind at ```$(document)``` event)
- removed most of key codes
- highlighting path logic
- wrapper element
- other small modifications(indentations, button class, unnecessary codes ... )

thank you.

app/assets/javascripts/project_find_file.js.coffee
```diff
@@ -0,0 +1,121 @@
...
+
+    @element.find(".tree-content-holder .tree-table").on 'click', (e) ->
+      if (e.target.nodeName != "A")
+        path = $('.tree-item-file-name a', this).attr('href')
+        if path then location.href = path
+
+  # find file
+  findFile: ->
+    searchTxt = @inputElement.val()
+    result = if searchTxt then fuzzaldrinPlus.filter(@filePaths, searchTxt) else @filePaths
```
DouweM: When `searchTxt` equals `""`, it will still be true and we enter fuzzAldrin. Please check if `searchTxt.length > 0`

app/assets/javascripts/project_find_file.js.coffee
```diff
@@ -0,0 +1,121 @@
...
+        @element.find('tr.tree-item').eq(0).addClass("selected").focus()
+
+    @element.find(".tree-content-holder .tree-table").on 'click', (e) ->
+      if (e.target.nodeName != "A")
+        path = $('.tree-item-file-name a', this).attr('href')
+        if path then location.href = path
+
+  # find file
+  findFile: ->
+    searchTxt = @inputElement.val()
```
DouweM: `searchTxt` is not much shorter than `searchText`, please use full variable names :)

app/assets/javascripts/project_find_file.js.coffee
```diff
@@ -0,0 +1,121 @@
...
+    $.ajax
+      url: url
+      method: "get"
+      dataType: "json"
+      success: ((data) ->
+        @element.find(".loading").hide()
+        @filePaths = data
+        @findFile()
+        @element.find('#files-slider tr.tree-item').eq(0).addClass("selected").focus()
+      ).bind(this)
```
DouweM: You should be able to use CoffeeScript's `=>` instead of `bind(this)`

app/assets/javascripts/project_find_file.js.coffee
```diff
@@ -0,0 +1,121 @@
...
+  load: (url) ->
+    $.ajax
+      url: url
+      method: "get"
+      dataType: "json"
+      success: ((data) ->
+        @element.find(".loading").hide()
+        @filePaths = data
+        @findFile()
+        @element.find('#files-slider tr.tree-item').eq(0).addClass("selected").focus()
```
DouweM: Can we not use ids on elements, but only classes instead?

app/assets/javascripts/project_find_file.js.coffee
```diff
@@ -0,0 +1,121 @@
...
+      ).bind(this)
+
+  # render result
+  renderList: (filePaths, searchTxt) ->
+    $(".tree-table > tbody").empty()
```
DouweM: Please find this within `@element` or something.

app/assets/javascripts/project_find_file.js.coffee
```diff
@@ -0,0 +1,121 @@
...
+
+    for filePath, i in filePaths
+      break if i > 19
```
DouweM: So if `i == 20`? :) That may be a little clearer.

app/assets/javascripts/project_find_file.js.coffee
```diff
@@ -0,0 +1,121 @@
...
+      highlightFilePath = filePath
+
+      if searchTxt
```
DouweM: `searchTxt.length > 0`

app/assets/javascripts/project_find_file.js.coffee
```diff
@@ -0,0 +1,121 @@
...
+        matches = fuzzaldrinPlus.match(filePath, searchTxt)
+        highlightFilePath = highlighter filePath, matches
+
+      blobItemUrl = "#{@options.blobUrlTemplate}/#{filePath}"
+      html = @makeHtml highlightFilePath, blobItemUrl
+      $(".tree-table > tbody").append(html)
```
DouweM: Please use `@element.find` or similar.

app/assets/javascripts/project_find_file.js.coffee
```diff
@@ -0,0 +1,121 @@
...
+
+      matchedChars.push(text[matchIndex])
+      lastIndex = matchIndex + 1
+
+    highlightText += matchedChars.join('').bold() if matchedChars.length
+    highlightText += text.substring(lastIndex)
+
+  # make tbody row html
+  makeHtml: (filePath, blobItemUrl) ->
+    return $("<tr class='tree-item'><td class='tree-item-file-name'><i class='fa fa-file-text-o fa-fw'></i><span class='str-truncated'><a href=#{blobItemUrl}>#{filePath}</a></span></td></tr>")
```
DouweM: To prevent XSS when `blobItemUrl` or `filePath` contain something that looks like HTML/JS, Build the `span` element like you do now, and then create the link like `$('<a>').attr('href', blobItemUrl).text(filePath)` and add it to the span using jQuery.

DouweM: Also, no need to explicitly return in CoffeeScript :)

app/assets/javascripts/project_find_file.js.coffee
```diff
@@ -0,0 +1,121 @@
...
+    @selectRow "UP"
+
+  selectRowDown: =>
+    @selectRow "DOWN"
+
+  goToTree: =>
+    location.href = @options.treeUrl
+
+  goToBlob: =>
+    path = $('.tree-item.selected .tree-item-file-name a').attr('href')
```
DouweM: `@element.find` please!

app/assets/javascripts/project_find_file.js.coffee
```diff
@@ -0,0 +1,121 @@
...
+    if path then location.href = path
```
DouweM: `location.href = path if path` is more CoffeeScript-like.

app/assets/javascripts/shortcuts_find_file.js.coffee
```diff
@@ -0,0 +1,15 @@
+#= require shortcuts_navigation
+
+class @ShortcutsFindFile extends ShortcutsNavigation
+  constructor: (@projectFindFile) ->
+    super()
+    _oldStopCallback = Mousetrap.stopCallback
+    Mousetrap.stopCallback = (event, element, combo) =>
+      if combo == 'up' or combo == 'down' or combo == 'esc' or combo == 'enter'
+        return element != @projectFindFile.inputElement[0]
+      return _oldStopCallback(event, element, combo)
```
DouweM: Why do we need to override `Mousetrap.stopCallback`?

app/views/projects/tree/show.html.haml
```diff
@@ -3,12 +3,12 @@
 = content_for :meta_tags do
   - if current_user
     = auto_discovery_link_tag(:atom, namespace_project_commits_url(@project.namespace, @project, @ref, format: :atom, private_token: current_user.private_token), title: "#{@project.name}:#{@ref} commits")
-
 = render 'projects/last_push'
 
 - if can? current_user, :download_code, @project
   .tree-download-holder
     = render 'projects/repositories/download_archive', ref: @ref, btn_class: 'btn-group pull-right hidden-xs hidden-sm', split_button: true
+= render 'projects/find_file_link'
```
DouweM: We need a little bit of distance between the "Find File" and "Download zip" buttons. This works:

- Change `_find_file_link.html` to:

```haml
= link_to namespace_project_find_file_path(@project.namespace, @project, @ref), class: 'btn btn-grouped shortcuts-find-file', rel: 'nofollow' do
  = icon('search')
  %span Find File
```

- Change in `tree/show.html.haml` from:

```haml
- if can? current_user, :download_code, @project
  .tree-download-holder
    = render 'projects/repositories/download_archive', ref: @ref, btn_class: 'btn-group pull-right hidden-xs hidden-sm', split_button: true
= render 'projects/find_file_link'
```
to:
```haml
.pull-right
  = render 'projects/find_file_link'
  - if can? current_user, :download_code, @project
    = render 'projects/repositories/download_archive', ref: @ref, btn_class: 'hidden-xs hidden-sm btn-grouped', split_button: true
```

DouweM:
@koreamic I may seem very critical (and I am), but I tried it out locally and it works really, really well. Great job. I can't wait to have this in the next release :) I'm sure we can get it ready.

app/assets/javascripts/shortcuts_find_file.js.coffee
koreamic: @DouweM 
When focus in textbox, Mousetrap's shortcuts is not working.
I think that we want to fire shortcuts action at that time in find file page.
so I overrided `Mousetrap.stopCallback`
and added `event.preventDefault` for avoiding to move cusor in textbox.
(when up/down key pressed, cusor is moved to home/end)

koreamic:
@DouweM 
I have to thank you for your review and concern all the time.
I have learned many things.
I modified sources and commited as your review.
but If there are things to be modified, leave comments.

DouweM:
@koreamic You're amazing. I've got nothing, I'm merging this.

JobV:
This is awesome. Thanks @koreamic

DouweM:
@koreamic I notice you're not credited in the changelog! I added `(koreamic)` in https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/2327, let me know if you want that to have your full name instead.

Razer6:
:heart: t

stanhu:
Great work, @koreamic!

koreamic:
@DouweM @stanhu @Razer6 @randx @JobV @yms9654 @noveler17
Thanks a lot for your help & patience.
@DouweM `koreamic` is enough. thanks

randx:
@koreamic can you please provide us with your full name? We prefer not to use usernames since it can be different between websites and can produce unnecessary confusion.  cc @DouweM

randx:
@koreamic and thank you for getting this feature done! :sparkles:

koreamic:
@randx @DouweM 
Okay. My full name is `Kyungchul Shin`
Thank you.

randx:
@koreamic thank you! 

@DouweM I will update CHANGELOG

randx:
wow @stanhu already updated CHANGELOG :smile: :rocket:

tsigo:
@koreamic Curious, why did you use fuzzaldrin-plus instead of fuzzaldrin?

DouweM:
@tsigo That was at my suggestion actually. I had much better results in Atom's fuzzy finder with the "Use Alternate Scoring" option enabled, which switches from fuzzaldrin to fuzzaldrin-plus. It's now the default in Atom's fuzzy finder as well: https://github.com/atom/fuzzy-finder/commit/c004c7ec19c631fbc3d01c4fd4a24fb95f9b601f

tsigo:
Ok. Do we know which version it was? I'm trying to replace the minified vendor JS with non-minified versions (specifically so we know version info, among other things).

DouweM:
@tsigo Ah, good call. It doesn't really matter, the API hasn't changed, so you can just get the latest one.

johnleeleelee:
Ok. Do we know which version it was? I'm trying to replace the minified vendor JS with non-minified versions (specifically so we know version info, among other things).
[https://www.jqueenwatchwinders.co.uk/best-jqueen-watch-winders-for-rolex
](url)

PR #9892, opened 2015-12-06
Fix gitlab-ssl nginx config to work when multiple server_names are served over https

stetskevych:
In a situation when you have several HTTPS sites served by nginx, you cannot use $server_name in the redirect, because it will always resolve to the first server_name specified in the config.
$http_host works in this scenario.


jacobvosmaer:
Thanks @stetskevych . In what NGINX version did $http_host get introduced? Will it work with the NGIXN in Ubuntu 12.04?

stetskevych:
@jacobvosmaer The new config is currently in production on CentOS 6 (running nginx 1.0 from EPEL). Ubuntu 12.04 ships nginx 1.1. So it will definitely work.

jacobvosmaer:
@stetskevych great!

PR #9894, opened 2015-12-08
update nginx config file hint

darkrasid:
> I recreate this PR because previous one takes wrong branches.
> # 9882

When use gitlab with relative url, 
user should change location regular expression on nginx config file.

Because example of nginx config file contain this type regular expression.

```
^/[\w\.-]+/[\w\.-]+/(info/refs|git-upload-pack|git-receive-pack)$
```

it will catch some URL like below.

```
/project/repo.git/git-upload-pack
```

However if user set relative url, upper url should be

```
/relative_url/project/repo.git/git-upload-pack
```

so regular expression isn't able to catch.

It makes errors when cloing, pulling, pushing or archiving.  
(ex. it will clone empty repository, even if the repository does not empty.)

I think it is better way that comment to user in example of nginx config file

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -49,6 +49,7 @@ v 8.3.0 (unreleased)
   - Fix deleting notes on a merge request diff
   - Display referenced merge request statuses in the issue description (Greg Smethells)
   - Implement new sidebar for issue and merge request pages
+  - Improve nginx config example file about gitlab with relative url. 
 
 v 8.2.3
   - Fix application settings cache not expiring after changes (Stan Hu)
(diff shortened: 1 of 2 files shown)


jacobvosmaer:
Thanks @darkrasid . I think this is being obsoleted by https://gitlab.com/gitlab-org/gitlab-workhorse/merge_requests/12 (relative URL support in gitlab-workhorse) and https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/2073 (move application-specific subrouting from NGIXN to gitlab-workhorse)

darkrasid:
Ok, I understand, thank you for your comment.

I will close this PR.

PR #9897, opened 2015-12-08
Fix inconsistency with no ssh key message

atomaka:
When a user has not added an SSH key, the alert message is inconsistent depending on if the project is empty or not.

Currently:

screen shot 2015-12-08 at 6 27 44 pm

This update will fix the message to match the show action by passing it as `content_for :flash_message`.  The result will be a more consistent look that has the additional space and appropriate link styling matching the non-empty project display.

screen shot 2015-12-08 at 6 26 41 pm

### DIFF
--- a/app/views/projects/empty.html.haml
+++ b/app/views/projects/empty.html.haml
@@ -1,4 +1,4 @@
-.alert_holder
+= content_for :flash_message do
   - if current_user && can?(current_user, :download_code, @project)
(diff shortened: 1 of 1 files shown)


atomaka:
Looks like master has not been mirrored from the GitLab repository.

randx:
@atomaka looks good. Thanks

PR #9918, opened 2015-12-17
Update deploy_key_multiple_projects.md : specify content-type in curl command

jle64:
The '--header "Content-Type: application/json"' was needed for the command to work for me.


stanhu:
Seems right to me. We should change the API to default to `application/json`, though. @axilleas, any comments?

axilleas:
Yeap, this is correct and it is [included in my MR](https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/681/diffs#diff-3) for the API clean-up. I will finish it this month, so thanks a lot @gentledevil but I'm gonna close this PR :/

PR #9924, opened 2015-12-21
allowing filtering of team city build events

glasspelican:


stanhu:
This change will cause merge request creations and new tag pushes to be sent to the TeamCity service, which doesn't appear to handle that data. I don't think that's what your intending here.

stanhu:
Closing due to inactivity.

PR #9932, opened 2015-12-28
Use ruby 2.1.8

pmq20:
Ruby 2.1.8 contains a security fix of CVE-2015-7551 which makes it more preferred. 

cf. https://www.ruby-lang.org/en/news/2015/12/16/ruby-2-1-8-released/


jvanbaarsen:
I think this should be merged in, but some work might already be done on upgrading ruby. /cc @DouweM

DouweM:
@jvanbaarsen We're planning to upgrade tot Ruby 2.2 for 8.5 (https://gitlab.com/gitlab-org/gitlab-ce/issues/3340), but in the mean time we should be on the latest 2.1. Build is failing though.

@pmq20 Would you mind retrying the build? Thanks!

stanhu:
Closing because GitLab has been updated to Ruby 2.2.4 now.

Issue #9637, opened 2015-09-16
Asana service is broken!


nalilord:
The Asana service is broken,

from the log:

Started GET "/name/repo/services/asana/test" for 127.0.0.1 at 2015-09-16 10:59:34 +0200
Processing by Projects::ServicesController#test as HTML
  Parameters: {"namespace_id"=>"name", "project_id"=>"repo", "id"=>"asana"}
Completed 500 Internal Server Error in 945ms

ActiveResource::ResourceNotFound (Failed.  Response code = 404.  Response message = Not Found.):
  app/models/project_services/asana_service.rb:112:in `block in check_commit'
  app/models/project_services/asana_service.rb:111:in `each'
  app/models/project_services/asana_service.rb:111:in `check_commit'
  app/models/project_services/asana_service.rb:94:in `block in execute'
  app/models/project_services/asana_service.rb:93:in `each'
  app/models/project_services/asana_service.rb:93:in `execute'
  app/controllers/projects/services_controller.rb:41:in `test'

I also tried to update the asana gem from 0.0.6 to the latest 0.3.0 but this doesn't work either, i guess they changed something essential, log output:

Started GET "/name/repo/services/asana/test" for 127.0.0.1 at 2015-09-16 15:15:57 +0200
Processing by Projects::ServicesController#test as HTML
  Parameters: {"namespace_id"=>"name", "project_id"=>"repo", "id"=>"asana"}
Completed 500 Internal Server Error in 40ms (ActiveRecord: 5.1ms)

NoMethodError (undefined method `configure' for Asana:Module):
  app/models/project_services/asana_service.rb:75:in `execute'
  app/controllers/projects/services_controller.rb:41:in `test'

So I leave it to you guys ;)

stanhu:
@j0k3r, do you know anything about this?

j0k3r:
1. Which version of Gitlab are you using?
1. Did you made any changes in the `asana_service.rb` ? Because reported lines aren't the same as in the repo.
1. Does the token is valid for you Asana project?
1. Does the user associated to this token has access to the target project in Asana?

As a side note, we are using Gitlab 7.14.1 and every thing works great.

nalilord:
Jep, I changed the message for the comment, but that's not the reason I believe, I also tried it with the official version, it's the same:

Started GET "/name/repo/services/asana/test" for 127.0.0.1 at 2015-09-16 17:09:16 +0200
Processing by Projects::ServicesController#test as HTML
[...]

PR #9937, opened 2015-12-29
[Bugfix] Change module to work with new asana api client version

wzrdtales:
# Description

These are the changes needed to make the asana module work with the updated asana gem again.
# Needed user action

With the upgrade the user now needs to use _Personal Access Tokens_ instead of _API Keys_. Thus the user also needs to replace all his _API Keys_ with _Personal Access Tokens_

fixes https://github.com/gitlabhq/gitlabhq/issues/9637

### DIFF
--- a/app/models/project_services/asana_service.rb
+++ b/app/models/project_services/asana_service.rb
@@ -72,7 +72,7 @@ automatically inspected. Leave blank to include all branches.'
   def execute(data)
     return unless supported_events.include?(data[:object_kind])
 
-    asana_client = Asana::Client.new do |c|
+    $asana_client = Asana::Client.new do |c|
       c.authentication :access_token, api_key
     end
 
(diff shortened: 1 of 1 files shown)


wzrdtales:
Whats even more useless then "the assignment" here is the message of robocop...

"app/models/project_services/asana_service.rb:75:5: W: Useless assignment to variable - asana_client."

As of http://www.rubydoc.info/github/bbatsov/rubocop/Rubocop/Cop/Lint/UselessAssignment this would be an unused variable, which this clearly is not. Either this is buggy or there is something I don't know about ruby. This could be as I'm not used to ruby, so please advice or just adjust.

app/models/project_services/asana_service.rb
```diff
@@ -72,8 +72,8 @@ def supported_events
   def execute(data)
     return unless supported_events.include?(data[:object_kind])
 
-    Asana.configure do |client|
-      client.api_key = api_key
+    $asana_client = Asana::Client.new do |c|
```
stanhu: You may want to change this to `@asana_client`.

stanhu:
Also see: https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/2111

That MR includes more changes that don't necessarily have to do with fixing Asana, but it does include specs that would help detect the breaking changes earlier. I think we'd want a fix that has your changes plus the basic specs.

stanhu:
Thanks for the PR. 8d5140502e8da08fde34f02db258d938b61eb83b was merged since it had additional documentation and tests and will be shipped in GitLab 8.3.3, so closing this for now.

PR #9938, opened 2015-12-30
Hotfix note mail with notification

huacnlee:
Today I found some error log:

``` bash
Email can not be processed: undefined method `note_note_email' for Notify:Class
```

When I reply a Issue comment by Email reply, It will create a new Note into database but not relation with the Issue.  

---

In current 8-3-stable (even in master version), `Emails::Notes` invoked `SentNotification.record`, that will store a bad `noteable` value.

See this file: https://github.com/gitlabhq/gitlabhq/blob/8-0-stable/app/mailers/emails/notes.rb#L46

@YorickPeterse

### DIFF
--- a/app/mailers/emails/notes.rb
+++ b/app/mailers/emails/notes.rb
@@ -48,7 +48,7 @@ module Emails
 
       yield
 
-      SentNotification.record(@note, recipient_id, reply_key)
+      SentNotification.record_note(@note, recipient_id, reply_key)
     end
   end
 end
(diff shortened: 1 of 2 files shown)


YorickPeterse:
I'm not sure how complex it would be, but is it possible to add a test for this case?

huacnlee:
Add a test for this case.

DouweM:
cc @tsigo This should go in an 8.3 patch release.

Razer6:
@DouweM @tsigo This was merged into `8-3-stable` - needs to be cherry-picked into `master`

DouweM:
@Razer6 Whoops, didn't verify the target branch.

tsigo:
Opened https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/2288 to pick into `master`.

PR #9952, opened 2016-01-06
Include the user_id in user_*_team system hooks.

tbeadle:
This fixes an issue where the user_id is not included in the data for
user_add_to_team and user_remove_from_team system hooks.  The documentation
already states that the user_id should be included.


randx:
:+1:

Razer6:
@tbeadle Can you please also update the documentation?

PR #9819, opened 2015-11-10
Replace CoffeeScript block into JavaScript in Views.

huacnlee:
Use CoffeeScript in View will compile again and again in each requests, because there have variables, Rails view can't caching them.

``` erb
:coffeescript
  projectId = #{@project.id}
```

By my benchmark test for `shared/issuable/_context` render time:
- CoffeeScript: `190ms`
- JavaScript: `19.7ms`

I have enabled all caching config like production environment.

PR #9960, opened 2016-01-11
use JavaScript instead of CoffeeScript in Views for master

huacnlee:
The reason #9819

@randx 

We need make sure the next changes not happen this case.

Use JavaScript in View not complex than CoffeeScript, and it faster, why not use that?

### DIFF
--- a/app/views/projects/_home_panel.html.haml
+++ b/app/views/projects/_home_panel.html.haml
@@ -52,5 +52,5 @@
 
     = render 'projects/buttons/notifications'
 
-:coffeescript
-  new Star()
+:javascript
+  new Star();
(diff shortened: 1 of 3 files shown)


randx:
cc @vsizov @tsigo @DouweM

DouweM:
@huacnlee Very good point, thanks! I will try to keep this in mind.

vsizov:
:+1:

tsigo:
Yep, this is a performance thing we need to all be aware of.

That said I don't think we should have _any_ inline Javascript, but that's another battle. :smile:

huacnlee:
https://github.com/gitlabhq/gitlabhq/pull/9959

Please merge #9959 , I need that, our team online GitLab service use 8-3-stable branch

tsigo:
Opened https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/2380 to remove the filters entirely so we don't have to worry about policing this from future contributors.

huacnlee:
@tsigo :+1:

PR #9977, opened 2016-01-15
update file path regex for unicode letters

yms9654:
When I upload a file, there is a issue if filename hava a unicode letters.(ex: korean or japanese or chinese characters).
It can't pass file_path_regex.
`[a-zA-Z0-9_\-\.\/]`

So how about this?
`[\p{L}\p{N}_\-\.\/]`
`\p{L}` means unicode letter and `\p{N}` is number.
http://ruby-doc.org/core-2.1.7/Regexp.html#class-Regexp-label-Character+Properties

### DIFF
--- a/lib/gitlab/regex.rb
+++ b/lib/gitlab/regex.rb
@@ -52,7 +52,7 @@ module Gitlab
     end
 
     def file_path_regex
-      @file_path_regex ||= /\A[a-zA-Z0-9_\-\.\/]*\z/.freeze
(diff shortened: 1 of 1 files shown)


connorshea:

If you'd like to get this merged, I encourage you to open this as a Merge Request on [the GitLab.com project](https://gitlab.com/gitlab-org/gitlab-ce). Feel free to ping me (@connorshea) and I'll get the right person assigned to it.

PR #9978, opened 2016-01-15
Update github.md

fantasywind:
add default callback URL path

### DIFF
--- a/doc/integration/github.md
+++ b/doc/integration/github.md
@@ -18,6 +18,7 @@ GitHub will generate an application ID and secret key for you to use.
     - Homepage URL: The URL to your GitLab installation. 'https://gitlab.company.com'
     - Application description: Fill this in if you wish.
     - Authorization callback URL: 'https://gitlab.company.com/'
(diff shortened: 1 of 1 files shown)


doc/integration/github.md
```diff
@@ -18,6 +18,7 @@ GitHub will generate an application ID and secret key for you to use.
     - Homepage URL: The URL to your GitLab installation. 'https://gitlab.company.com'
     - Application description: Fill this in if you wish.
     - Authorization callback URL: 'https://gitlab.company.com/'
+    - If install from source, default callback URL is '${YOUR_DOMAIN}/import/github/callback'
```
stanhu: Doesn't this apply to omnibus installations as well?

fantasywind: I install from source only, cannot know about omnibus path structure, sor.
If structure is same with source installation, I can fix it!

axilleas: Yes, it's the same in Omnibus as well. Remove `If install from source` and we can merge :)

axilleas:
Cool, thanks @fantasywind

PR #9997, opened 2016-01-22
Pass username to Piwik analytics when Piwik is enabled

Senorsen:
Pass username to Piwik, through `setUserId` parameter.


Senorsen:
Oh why did the test fail?

PR #10016, opened 2016-02-04
Fix gitlab init script

TonyTsangHK:
Actual command (cd $gitlab_workhorse_dir) should not be run during variable declaration.

### DIFF
--- a/lib/support/init.d/gitlab
+++ b/lib/support/init.d/gitlab
@@ -47,8 +47,6 @@ shell_path="/bin/bash"
 # Read configuration variable file if it is present
 test -f /etc/default/gitlab && . /etc/default/gitlab
 
-cd $gitlab_workhorse_dir
-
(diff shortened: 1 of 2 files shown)


jacobvosmaer:
@TonyTsangHK why do you think this change should be made?

TonyTsangHK:
If $app_root/../gitlab-workhorse does not exists, cd to it will generate unnecessary error message.

It will always happen when using custom app_root, like me.

Just a minor fix to avoid cd error message, the init script will work even without it.

jacobvosmaer:
Thanks @TonyTsangHK . I think I prefer the solution in https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/2697

TonyTsangHK:
That will do trick too.

I am closing the PR.

jacobvosmaer:
Thanks for this PR anyway @TonyTsangHK !

PR #10017, opened 2016-02-04
6 2 stable

itisguptak:
Need for deploy


Razer6:
All changes of 6-2-stable are included in master.

PR #10023, opened 2016-02-09
Change default smpt port to 465 (default smtps port)

bozaro:
Current default value 456 is confusing. It differ with default SMTPS port number 465 only by digits order.
I think it's good idea to change this value to some correct SMTP port number (25 or 465).


rymai:
This looks good to me, thanks @bozaro.

@DouweM I let you review/confirm/merge. Thanks!

PR #10054, opened 2016-02-24
Add scoverage (Scala) example regex for finding test coverage

jasperdenkers:
Adds an example regex on the `Project Setting > Continous Integration > Test coverage parsing` page for Scala test coverage measuring with scoverage.

See below for how a scoverage output would look like when e.g. running `sbt clean coverage test coverageReport`:

image


jasperdenkers:
Now covered in https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/2984.

PR #10077, opened 2016-03-11
fix #10076

### LINKED ISSUE #10076: Incorrect parameter's description [docs]
[In projects docs](https://github.com/gitlabhq/gitlabhq/blob/master/doc/api/projects.md) it's said that parameter name should be `visibility`, but I get 500 error if set this parameter. But if set it to `visibility_level` everything is ok.

lgg:
fix visibility parameter description


littleguga:
Unconfirmed see #10076 for more info.

PR #10100, opened 2016-03-23
Fix order of steps to prevent PostgreSQL errors when running migration

rodrigomanhaes:
In the docs about upgrading 8.5 to 8.6, the step that explains PostgreSQL requirements for Gitlab 8.6 should come before the migration to prevent errors.


stanhu:
Looks right to me. @YorickPeterse, thoughts?

rymai:
Thanks @rodrigomanhaes, I've reopened this at https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/3355.

PR #10118, opened 2016-04-06
Changed the argument of not_found for 'unprotect'

yasserhussain1110:
The error message I got when I tried to unprotect a non-existing branch didn't seem right. This pull request will fix that.


rymai:
Thanks @yasserhussain1110!

PR #10140, opened 2016-04-15
fix required gitlab-shell version.

lurdan:
Under the procedure, I've warned about gitlab-shell version, which should be 2.6.12 (maybe some minor revisions introduces that).


rymai:
Thanks @lurdan. Indeed, it was changed in 8.6.3 (it's in the `8-6-stable` and `8-6-stable-ee` branch but not in the `master` branch)!

PR #10153, opened 2016-04-22
Update build triggers API documentation

IgnoredAmbience:
- Return body example missing for delete method.
- Fixed incorrect parameter description.


axilleas:
Thanks!

PR #10311, opened 2016-04-27
Don’t read otp_secret_encryption_key from hardcoded path in models/user

jirutka:
Variable `Gitlab::Application.config.secret_key_base` is set in `config/initializers/secret_token.rb`. It’s very bad practice to use hard-coded paths inside an application and really unnecessary in this case.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -85,7 +85,7 @@ class User < ActiveRecord::Base
   default_value_for :theme_id, gitlab_config.default_theme
 
   devise :two_factor_authenticatable,
-         otp_secret_encryption_key: File.read(Rails.root.join('.secret')).chomp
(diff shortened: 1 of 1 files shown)


rymai:
Thanks @jirutka!

Looks good to me, @tsigo what do you think since you added 2FA support in 66bc758e.

@jirutka Could you open a new MR at https://gitlab.com/gitlab-org/gitlab-ce/merge_requests so that we're tests still pass? Thanks in advance.

tsigo:
LGTM.

jirutka:
Okay, I did it, but it’d be much better if you fix your tests on GitHub…

https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/4044

rymai:
@jirutka Thanks!

PR #10324, opened 2016-05-04
Add support for MathJax with AsciiDoc

jirutka:
This pull requests adds optional (disabled by default) support for MathJax for typesetting of math used in AsciiDoc files with possibility to extending for other markups.

MathJax library is quite big, so it’s not integrated into asset pipeline, but instead loaded from CDN only when actually needed. The detection for AsciiDoc is very simple, because Asciidoctor has a built-in macro for STEM (i.e. math) that must be explicitly enabled for the document using attribute `stem` (i.e. `:stem:`).

It should work everywhere where AsciiDoc works, I’ve tested it in Readme, Files (including preview when editing a file), Wiki and Snippets. It works with turbolinks well.

Related to https://gitlab.com/gitlab-org/gitlab-ce/issues/13690 and #5280.

### DIFF
--- a/app/assets/javascripts/blob/edit_blob.js.coffee
+++ b/app/assets/javascripts/blob/edit_blob.js.coffee
@@ -34,6 +34,7 @@ class @EditBlob
       , (response) ->
         currentPane.empty().append response
         currentPane.syntaxHighlight()
+        $(document).trigger('mathjax:typeset')
 
     else
       @editor.focus()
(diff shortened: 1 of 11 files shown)


jirutka:
@rymai ping

connorshea:
Closing since this was moved over to https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/4333

PR #10338, opened 2016-05-16
typo fix: # Disable all components except Redis

donhui:
typo fix: # Disable all components except Redis


tsigo:
@donhui Thanks!

PR #10365, opened 2016-07-01
update Gemfile

bogdanvlviv:
1. replace `"` to `'` in Gemfile.
2. change `require: nil` to `require: false`.
   `require: nil` and `require: false` get same result, but i think we shouldn't write it differently in other places.
3. set version for gems.
   better gem's version control for prevent dependency errors.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,4 +1,4 @@
-source "https://rubygems.org"
+source 'https://rubygems.org'
 
 gem 'rails', '4.2.6'
 gem 'rails-deprecated_sanitizer', '~> 1.0.3'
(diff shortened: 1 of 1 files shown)


connorshea:
Thanks for the PR! Because we want to "eat our own dogfood" and keep all our issues and code review in one place, we're only accepting merge requests on GitLab.com.

You can log in with your GitHub account and submit an MR to the project here: https://gitlab.com/gitlab-org/gitlab-ce/merge_requests

bogdanvlviv:
Got it. https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/5078

PR #10421, opened 2017-03-28
Fix Block and Unblock api method

LordS4me:
Seems api method for Block and Unblock api is PUT and not POST


rymai:
@LordS4me Thanks. These endpoints has changed in the v4 API: they're now using `POST`. See https://docs.gitlab.com/ce/api/v3_to_v4.html#9-0.

PR #10424, opened 2017-04-09
Add gitlab:ldap:generate_avatars rake task

brigriffin:
Add gitlab:ldap:generate_avatars rake task to use and generate avatar file from LDAP for all users

See issue #14459

### DIFF
--- a/lib/gitlab/ldap/adapter.rb
+++ b/lib/gitlab/ldap/adapter.rb
@@ -101,7 +101,7 @@ module Gitlab
       end
 
       def user_attributes
-        %W(#{config.uid} cn mail dn)
+        %W(#{config.uid} cn mail dn jpegPhoto)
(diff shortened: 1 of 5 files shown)


dblessing:
Hi @brigriffin 

Thanks for your contribution. I would like to see this integration within the `LdapSync` flow rather than via a one-off Rake task. Do you think you can rework this to update user attributes during sign in and during the `LdapSync` (not to be confused with `LdapGroupSync`)? I will close this for now but if you update it or open a new pull request, please ping me and we can re-open it and review it again. Thanks again for your contribution.

brigriffin:
Thanks @dblessing for your suggestion. I will first need to dig into the `LdapSync` flow in order to see how it works and will give it a shot as soon as I have some time.

PR #10472, opened 2018-02-06
Add skip_password_expiration boolean when doing PUT on users

eLvErDe:
Hello,

We're synchronizing users's password using admin API key and recent GitLab CE releases broke this behavior by adding an "already-expired-password" feature when a user's password is being changed throught the API. That leads to an unexpected password change prompt when the users logs in.

Altough I understand this feature I think our usage makes sense too and both should be allowed.

This pull request adds a boolean to override password expiration to the API.

### DIFF
--- a/doc/api/users.md
+++ b/doc/api/users.md
@@ -298,6 +298,7 @@ Parameters:
 - `admin` (optional)            - User is admin - true or false (default)
 - `can_create_group` (optional) - User can create groups - true or false
 - `skip_reconfirmation` (optional) - Skip reconfirmation - true or false (default)
+- `skip_password_expiration` (optional) - Do not ask user to change his password at next login - true or false (default)
 - `external` (optional)         - Flags the user as external - true or false(default)
 - `avatar` (optional)           - Image file for user's avatar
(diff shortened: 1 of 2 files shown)


godfat:
I am sorry, but we're not using GitHub for this. Could you please create a merge request at https://gitlab.com/gitlab-org/gitlab-ce/merge_requests instead?

eLvErDe:
Cannot create the pull request on your GitLab, stuck at "Forking in progress" for an hour

godfat:
I am sorry that happened :( I'll poke around and see if we have some infrastructure issue. In the mean time, perhaps you could try to clone it first, and create an empty project, and push it to the empty project?

Sorry for the inconvenience, and thank you for the contribution!

eLvErDe:
I'll come back later and check if it finally ends, or even tomorrow.

eLvErDe:
Here we go :)

https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/16938

godfat:
Let's follow up there, thank you!

PR #10485, opened 2018-04-18
enabled word breaking of card title

oimyounis:
Enabled word breaking of card title as it gets overlapped by the assignee's icon when its a single long line with no spaces.


QuangThong81:
Mau xanh luc

PR #10493, opened 2018-08-25
Update 20160610201627_migrate_users_notification_level.rb

AhmedAliGokhy:
Hey i hope all is well,
While i reading Gitlab migrations files to learn what is the best practice to migrate new changes on old schema i noticed that in this file 20160610201627_migrate_users_notification_level DOWNTIME is duplicated without any reasons !! Why ?!

Thank you for taking the time to contribute back to GitLab!

Please open a merge request [on GitLab.com](https://gitlab.com/gitlab-org/gitlab-ce/merge_requests), we look forward to reviewing your contribution! You can log into GitLab.com using your GitHub account.

### DIFF
--- a/db/migrate/20160610201627_migrate_users_notification_level.rb
+++ b/db/migrate/20160610201627_migrate_users_notification_level.rb
@@ -4,8 +4,6 @@ class MigrateUsersNotificationLevel < ActiveRecord::Migration
   # Migrates only users who changed their default notification level :participating
   # creating a new record on notification settings table
 
-  DOWNTIME = false
-
   def up
     execute(%Q{
       INSERT INTO notification_settings


mkozono:
Hi @AhmedAliGokhy, this repo is only maintained on GitLab.com now. Please submit merge requests there https://gitlab.com/gitlab-org/gitlab-ce/merge_requests

However, I can say that if you look at the last commit to this file (I will leave it to you to find, for your education), you will see the commit message "Fix migration without DOWNTIME clause specified" and they just added the line to the top. This was certainly a mistake, and it is clear from the message that they "had" to add it. The reason is probably that a Rubocop style cop was added to ensure that the constant exists at the top of the migration class.

It doesn't hurt the migration for it to be defined twice, but anyway, that's what happened.

mkozono:
The first thing to do when you want to know "why", is to "git blame" the file. https://stackoverflow.com/questions/31203001/what-does-git-blame-do

Good luck!

AhmedAliGokhy:
Thanks for your consideration.

